PR #4, opened 2016-08-22
Sonatype

C0urante:


criccomini:
This all looks good. One nit about the version location.

build.gradle
```diff
@@ -3,23 +3,102 @@ apply plugin: 'idea'
...
-version = '0.2'
 
-jar {
-  manifest {
-    attributes  'Implementation-Title': 'Kafka-Connect-BigQuery',
-                'Implementation-Version': version
-  }
+group = 'com.wepay'
+archivesBaseName = 'kafka-connect-bigquery'
+version = '0.2.1'
```
criccomini: Would be good to have this in a gradle.properties file locally. See [here](http://stackoverflow.com/questions/17262856/gradle-how-to-set-the-project-name-group-version-plus-source-targetcompatibi).

PR #10, opened 2016-08-23
Spell WePay correctly.

skyzyx:
* Spell _WePay_ correctly. We need to be on-brand at all times.
* Ensure the Bash interpreter is called uniformly across all shell files.
* Remove all trailing whitespace.


C0urante:
Looks good to me. Merged.

PR #29, opened 2016-08-26
Adjust the upload process to reflect the new module structure

C0urante:


build.gradle
```diff
@@ -1,15 +1,21 @@
-apply plugin: 'java'
+// BEGIN ALL PROJECTS //
+allprojects {
+    apply plugin: 'java'
+    version = '0.3.0-SNAPSHOT'
```
criccomini: I think you can set `version` at the root level.

Issue #24, opened 2016-08-24
Use docs from Kafka connect structs


criccomini:
We should use the [doc](https://kafka.apache.org/0100/javadoc/org/apache/kafka/connect/data/Schema.html#doc()) from Kafka connect's structs to set the documentation string for tables and fields in BigQuery.

criccomini:
top-level struct’s doc is used as the description for the table itself, and each subsequent field’s doc is used as the description for the BigQuery field?

PR #32, opened 2016-08-26
Issue #24

mtagle:
BQ tables will now have descriptions matching the kafka schema documentation.
Added verification test.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
@@ -56,9 +56,12 @@ public class SchemaManager {
     bigQuery.update(constructTableInfo(table, kafkaConnectSchema));
   }
 
-  private TableInfo constructTableInfo(TableId table, Schema kafkaConnectSchema) {
+  // package private for testing.
+  TableInfo constructTableInfo(TableId table, Schema kafkaConnectSchema) {
     com.google.cloud.bigquery.Schema bigQuerySchema =
         schemaConverter.convertSchema(kafkaConnectSchema);
-    return TableInfo.of(table, StandardTableDefinition.of(bigQuerySchema));
+    return TableInfo.builder(table, StandardTableDefinition.of(bigQuerySchema))
+      .description(kafkaConnectSchema.doc())
+      .build();
   }
 }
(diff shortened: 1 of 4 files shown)


C0urante:
I think @criccomini was also hoping that each [description()](http://googlecloudplatform.github.io/gcloud-java/0.2.8/apidocs/com/google/cloud/bigquery/Field.html#description()) field of each of the resulting BigQuery Schema's Fields could be populated by the [doc()](http://docs.confluent.io/3.0.0/connect/javadocs/org/apache/kafka/connect/data/Schema.html#doc--) field of each of the corresponding Kafka Connect Schemas (since each Kafka Connect Field comes with its own Schema).

In order to accomplish that, it'd probably make more sense to make this change in the BigQuerySchemaConverter class instead of the SchemaManager, with the top-level Kafka Connect Struct's doc() field being used for the resulting BigQuery Table's description field.

mtagle:
ack on the descriptions of the individual fields.

As for the description of the BigQuery tables, those aren't part of the BQ schemas, so I can't use the schema converters to get the kafka docs into the BQ table descriptions.

mtagle:
Brief thought about the general unpleasantness of having the table documentation conversion happen in SchemaManager rather than SchemaConverter:
One way to get around this would be to have our SchemaConvert actually convert from Kafka Schemas to BQ TableInfos.  This wouldn't have to change the code that much because converting from BQ Schema to TableInfo is just a couple of lines that we can just tack onto the end of the Schema Conversion. I think this would make the code more confusing, though, and I'm not sure the trade-off is worth it.  I'm open to opinions though.

C0urante:
Did some thinking about it and I like your idea of converting from KC Schema to BQ TableInfo. I think if changes have to be made in the future, it's more likely they'll occur with the SchemaManager and less likely (like a *lot* less likely) with the SchemaConverter, so it would behoove us to keep from complicating the SchemaManager too much.

I also noticed that you have checks for null Field descriptions in the SchemaConverter but not in the SchemaManager. Have you tested what happens when a null description is set for a TableInfoBuilder? And either way, just to be safe, would you consider adding a unit test as a sanity check just in case Google breaks things in a future release?

mtagle:
Okay, after messing with this a bit, my idea of having the SchemaConverts convert to TableInfos doesn't quite work as well as I had hoped.

TableInfos have their BQ TableId in them, so to convert to a TableInfo you would need more than just the kafka schema: you'd also need to know where in BQ you wanted this schema to end up, which is awkward.

I thought maybe we could get around this by passing around TableInfo.Builders instead, containing schema and table documentation but no TableId, that you would later pass a TableId to before building into a full TableInfo, but you can't do that either. TableInfo.Builders **require** a TableId.  You have to pass one when creating the builder and google's code will verify that it's not null when that happens. Personally I think the idea that you can't put together a builder piecemeal is a poor design decision on google's part, but that's what we are working with.

I have the much smaller change of only setting the table documentation if it's not null though, that I will push here. The rest of the changes that I made will just remain on my box for now, until we make a decision about what we want to do.

I agree that setting the tableDocumentation in SchemaManager is messy, but after all this work for nothing I'm not convinced that alternatives will be better.

C0urante:
Yeah, I definitely wouldn't spend any more time investigating how to put the changes solely into the SchemaConverter. The complexity involved in that is a pretty convincing argument that it's just simpler to allow things to happen in the SchemaManager.

Looks good to me, merging.

PR #36, opened 2016-09-06
Fix for optional structs causing NPEs.

mtagle:


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java
```diff
@@ -72,6 +72,15 @@
 
   @SuppressWarnings("unchecked")
   private Object convertObject(Object kafkaConnectObject, Schema kafkaConnectSchema) {
+    if (kafkaConnectObject == null) {
+      if (kafkaConnectSchema.isOptional()){
+        // short circuit converting the object
+        return null;
+      } else {
+        return new ConversionConnectException(
```
mtagle: in practice, this will never happen, as far as I can tell, because kafka Structs won't allow you to put a null value in a field that isn't optional; but I felt the code looked cleaner and clearer this way.

criccomini:
LGTM!

Issue #34, opened 2016-09-06
Handle null values


criccomini:
Kafka log compacted topics can receive null value messages, which signify deletes for a given key. The connector should handle these messages in some way. Either ignore them or something.

criccomini:
Best idea that I have is to have KCBQ ignore hard deletes. The primary use case that we have for this is to handle deletions from [Debezium](http://debezium.io/), which sends a soft delete then a hard delete.

Perhaps in the future we might want some way to funnel delete messages into a separate table, or to even do real updates using

PR #38, opened 2016-09-07
Modify BigQuerySinkTask to be able to handle null messages.

mtagle:
This is needed because debezium sends null messages when it is deleting messages in kafka.
We do nothing because we do not want hard deletes in BQ.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -234,10 +234,13 @@ public class BigQuerySinkTask extends SinkTask {
     }
   }
 
-  private Map<TableId, List<SinkRecord>> partitionRecordsByTable(
-      Collection<SinkRecord> records) {
+  private Map<TableId, List<SinkRecord>> partitionRecordsByTable(Collection<SinkRecord> records) {
     Map<TableId, List<SinkRecord>> tableRecords = new HashMap<>();
     for (SinkRecord record : records) {
+      if (recordEmpty(record)) {
+        // ignore it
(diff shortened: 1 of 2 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -234,10 +234,13 @@ private RowToInsert getRecordRow(SinkRecord record) {
     }
   }
 
-  private Map<TableId, List<SinkRecord>> partitionRecordsByTable(
-      Collection<SinkRecord> records) {
+  private Map<TableId, List<SinkRecord>> partitionRecordsByTable(Collection<SinkRecord> records) {
     Map<TableId, List<SinkRecord>> tableRecords = new HashMap<>();
     for (SinkRecord record : records) {
+      if (recordEmpty(record)) {
+        // ignore it
```
criccomini: Could you add a DEBUG log line here, saying we're dropping a sink record?

criccomini:
LGTM.. once request for a log line

criccomini:
Fixes https://github.com/wepay/kafka-connect-bigquery/issues/34

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -234,10 +234,14 @@ private RowToInsert getRecordRow(SinkRecord record) {
...
+        logger.debug("ignoring empty record value for key: " + record.key().toString());
```
criccomini: Keys can be null in Kafka as well. It would be super weird to send a null key and null value, but it's possible.

PR #40, opened 2016-09-13
Di 448

mtagle:
also some style fixes that I had previously missed.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -152,8 +152,7 @@ public class BigQuerySinkTask extends SinkTask {
     @Override
     public Void call() throws InterruptedException {
       List<List<RowToInsert>> partitions = rowPartitioner.partition(rows);
-      bigQueryWriter.writeRows(table, partitions.get(0), topic, schemas);
-      for (List<RowToInsert> partition : partitions.subList(1, partitions.size())) {
+      for (List<RowToInsert> partition : partitions) {
         bigQueryWriter.writeRows(table, partition, topic, schemas);
       }
       updateAllPartitions(tablesToTopics.get(table), offsets);
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -88,8 +88,6 @@ import java.util.concurrent.Future;
 public class BigQuerySinkTask extends SinkTask {
   private static final Logger logger = LoggerFactory.getLogger(BigQuerySinkTask.class);
 
-  public static final long TABLE_WRITE_INTERVAL = 1000L;
-
   private final ExecutorService executorService = Executors.newCachedThreadPool();
 
   private final Object contextLock = new Object();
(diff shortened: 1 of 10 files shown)


mtagle:
so right now I have the retry logic for quota exceeded as separate from the retry logic for 500/503.  This is kind of confusing, but it seemed justified to me.  I'm open to opinions though. (Or Ideas on how I can unify them a bit)

also, technically 403 covers more than just quota exceeded errors.  Right now I'm just blindly assuming that any 403 should be retried and doing that. There are a couple of 403s that probably shouldn't be, but they seem like very edge casey cases: accessDenied, billingNotEnabled, blocked, and responseTooLarge (this shouldn't ever happen with what we're doing, as far as I know). There is also rateLimitExceeded, but I put that in the same bucket as quotaExceeded.

mtagle:
used this SO question to aid in figuring out how to tell quotaExceeded 403 from others: http://stackoverflow.com/questions/38573446/bigquery-decyphering-quota-exceeded-message (because of course this doesn't seem to be documented by google...)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -156,7 +154,6 @@ public Void call() throws InterruptedException {
       List<List<RowToInsert>> partitions = rowPartitioner.partition(rows);
       bigQueryWriter.writeRows(table, partitions.get(0), topic, schemas);
```
criccomini: Clean this loop

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/BigQueryWriter.java
```diff
@@ -123,10 +133,24 @@ public void writeRows(
         if (err.code() == INTERNAL_SERVICE_ERROR || err.code() == SERVICE_UNAVAILABLE) {
           // backend error: https://cloud.google.com/bigquery/troubleshooting-errors
           retryCount++;
+        } else if (err.code() == FORBIDDEN
+                   && err.error() != null
+                   && QUOTA_EXCEEDED_REASON.equals(err.error().reason())) {
+          // quota exceeded error
```
criccomini: Would be good to have at least a debug log here, so we know when we're hitting quotas.

Also would be good to have a metric to count retries.

criccomini:
LGTM just a couple of nits

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
mtagle: oops, ty

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/BigQueryWriter.java
```diff
@@ -72,6 +87,21 @@ public BigQueryWriter(int retries, long retryWaitMs, Metrics metrics) {
...
+    requestRetries.add(metrics.metricName("request-retries-avg",
+                                          MetricsConstants.groupName,
+                                          "The average number of retries per request"),
+                       new Avg());
+    requestRetries.add(metrics.metricName("request-retries-max",
+                                          MetricsConstants.groupName,
+                                          "The maximum number of retry attempts made for a single "
+                                          + "request"),
+                       new Max());
+    requestRetries.add(metrics.metricName("request-retires-count",
```
criccomini: retires -> retries

PR #42, opened 2016-09-27
Dynamic batch sizes

mtagle:


mtagle:
Some things we should decide:
I'm not sure if "partitioner" is really the right parlance anymore.  Maybe... RequestManager?
Should we deprecate the existing max batch size config and create a new config for the partitioner?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -303,10 +300,13 @@ private boolean recordEmpty(SinkRecord record) {
...
     int maxWriteSize = config.getInt(config.MAX_WRITE_CONFIG);
+    BigQueryWriter writer = getWriter();
     if (maxWriteSize == -1) {
-      return new SinglePartitioner<>();
+      return new SinglePartitioner(writer);
+    } else if (maxWriteSize == 0) {
+      return new DynamicPartitioner(writer);
     } else {
-      return new EqualPartitioner<>(maxWriteSize);
+      return new EqualPartitioner(writer, maxWriteSize);
```
criccomini: I'm in favor of just deleting equal partitioner and having either single or dynamic. Also in favor of eliminating maxWriteSize, and just having a config to define which partitioner to use single, or dynamic.

Thoughts?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/EqualPartitioner.java
```diff
@@ -18,70 +18,90 @@
  */
 
 
-import java.util.ArrayList;
-import java.util.Collections;
+import com.google.cloud.bigquery.BigQueryException;
```
criccomini: I want to nuke this file.

criccomini:
Agree we should rename partitioner. My vote is for _BatchWriter_.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/Partitioner.java
```diff
@@ -18,16 +18,28 @@
...
-   * @return A series of lists, which when combined in order is equal to the argument list.
+   * @param table The BigQuery table to write the rows to.
+   * @param elements The list of elements to write to BigQuery.
+   * @param topic The Kafka topic that the row data came from.
+   * @param schemas The unique Schemas for the row data.
+   * @throws BigQueryConnectException if we are unable to write to BigQuery
+   * @throws InterruptedException if writing is interrupted
    */
-  public List<List<E>> partition(List<E> elements);
+  void writeAll(TableId table, List<E> elements, String topic, Set<Schema> schemas)
```
criccomini: Since we're now writing the elements, I don't think that templating this makes sense anymore. Should just be a collection or iterable.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+    contSuccessCount++;
+    if (contSuccessCount >= CONT_SUCCESS_COUNT_BUMP) {
+      logger.debug("{} successful establishedWriteAlls in a row, increasing batchSize to {}",
+                   contSuccessCount, currentBatchSize);
+      increaseBatchSize();
+    }
+  }
+
+  private boolean isPartitioningError(BigQueryException exception) {
+    if (exception.code() == BAD_REQUEST_CODE
```
criccomini: Add a detailed note here about why we're doing this. It's a bit odd, and I could see myself forgetting.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+  private int currentBatchSize;
+  private boolean seeking;
+
+  private int contSuccessCount;
+
+  /**
+   * @param writer the {@link BigQueryWriter} that will do the writing.
+   */
+  public DynamicPartitioner(BigQueryWriter writer) {
+    this.writer = writer;
```
criccomini: Nit: Can just call the below constructor.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+        if (currentBatchSize == MAXIMUM_BATCH_SIZE) {
+          seeking = false; // case 3
+          logger.debug("Best batch size found (max): {}", currentBatchSize);
+          establishedWriteAll(table,
+                              elements.subList(currentIndex, elements.size()),
+                              topic,
+                              schemas);
+          return;
+        }
+        //  increase the batch size if: we actually did just test that batchSize
```
criccomini: Nit: double space

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+  /**
+   * writeAll intended for a generally stable currentBatchSize.
+   *
+   * <p>Write all the given elements to BigQuery in the smallest possible number of batches, as
+   * evenly as possible.
+   *
+   * <p>If we encounter a partitioning-related error, we will immediately lower the batch size and
+   * try again.
+   *
+   * <p>Every {@link #CONT_SUCCESS_COUNT_BUMP} calls, if there have been no errors, we will bump up
```
criccomini: CONT->CONST

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+    return false;
+  }
+
+  // for some reason when I run tests it's asking for these to be public?? I'm not sure why...
+  private void increaseBatchSize() {
+    currentBatchSize = Math.min(currentBatchSize * 2, MAXIMUM_BATCH_SIZE);
+  }
+
+  private void decreaseBatchSize() {
+    currentBatchSize = currentBatchSize / 2;
```
criccomini: Should enforce at least 1

    Math.max(1, currentBatchSize / 2);

criccomini: Or maybe we should puke if we're getting msg too large errors from Google with a batch size of 1?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+
+    contSuccessCount++;
+    if (contSuccessCount >= CONT_SUCCESS_COUNT_BUMP) {
+      logger.debug("{} successful establishedWriteAlls in a row, increasing batchSize to {}",
+                   contSuccessCount, currentBatchSize);
+      increaseBatchSize();
+    }
+  }
+
+  private boolean isPartitioningError(BigQueryException exception) {
```
criccomini: I think this method should also handle the "real" 403 errors that we *should* get from Google. I have a fear that some NOC admin will change the frontend max > 10mb, and then we'll start getting 403's instead of 400s, and Google will have no idea.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+          currentIndex = endIndex;
+        }
+      } catch (BigQueryException exception) {
+        if (isPartitioningError(exception)) {
+          // immediately decrease batch size and try again with remaining elements.
+          decreaseBatchSize();
+          contSuccessCount = 0;
+          logger.debug("Partition error during establishedWriteAll, reducing batch size to {}",
+                       currentBatchSize);
+          continue;
```
criccomini: No need for `continue`, right?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
criccomini: Some javadocs

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+        } else {
+          throw new BigQueryConnectException(
+              String.format("Failed to write to BigQuery table %s", table),
+              exception);
+        }
+      }
+    }
+
+    contSuccessCount++;
+    if (contSuccessCount >= CONT_SUCCESS_COUNT_BUMP) {
```
criccomini: Does `contSuccessCount` need to get reset in this block? If not, it seems like the batch size will get bumped on every subsequent write once contSuccessCount >= CONT_SUCCESS_COUNT_BUMP is reached.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+      try {
+        // very similar to equalPartitioner
+        // handle case where no partitioning is necessary:
+        if (elements.size() <= currentBatchSize) {
+          writer.writeRows(table, elements, topic, schemas);
+          currentIndex = elements.size();
+          return;
+        }
+
+        int numPartitions = (int)Math.ceil(elements.size() / (currentBatchSize * 1.0));
```
criccomini: Not sure how this impacts things, but one observation is that these variables are all calculated off of `elements.size()`. If there's a write failure due to size, and decreaseBatchSize is decreased, there will be a bit of skew going on, since currentIndex won't be zero, and you'll be changing batch size.

I think you might want to change this logic to calculate (at least) numPartitions based off of elements.size() - currentIndex, and think through how this impacts things.

criccomini:
Also, before merging, could you run this against a real workload that has large messages/triggers increase and decrease in batch size locally and just confirms that it behaves as expected.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
mtagle: Sounds good to me.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/Partitioner.java
mtagle: Can you elaborate on what you mean?

criccomini: I don't think that the `<E>` is needed anymore. Since the writer is actually used within the partitioner now, and the writer requires using `InsertAllRequest.RowToInsert`, I think it makes sense to just force `InsertAllRequest.RowToInsert` as the type.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: done

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: fixed

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
```diff
@@ -0,0 +1,256 @@
...
+   *
+   * <p>We start with the currentBatchSize and adjust up or down as necessary. Seek mode ends when
+   * one of the following occurs:
+   *   1. we get an error after at least 1 success. This is slightly inefficient if our
+   *      currentBatchSize is too big, but we will only make 1 extra request in that case.
+   *      Successes do not carry over from previous seekingWriteAll requests.
+   *   2. we write all the given elements in one batch request without error.
+   *   3. we write MAXIMUM_BATCH_SIZE elements in one batch request without error. (unlikely)
+   */
+  private void seekingWriteAll(TableId table,
```
mtagle: it's the same signature as writeAll though, which is documented in Partitioner

criccomini: Ah missed that.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: I mean cont as in continuous (how many successes in a row before we bump the batchSize up)

criccomini: Oh... I did not get that lol. :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: I think you are correct.  I will fix that.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: no; I'll remove it.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: oops, yes.  fixed.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: mm, alright.

mtagle: although, what do we expect the "real" 403s would look like? quotaExceeded errors?

criccomini: Ya, exactly.

Also, unrelated question: are we handling rateLimitExceeded anywhere in the code? IIRC, somewhere should be handling this (sleep a second then retry). I think the writer might be doing this? Just want to confirm.

mtagle: because right now I am catching quotaExceededs in BigQueryWriter because right now they mean you have too many requests per second (and that's where I do the pause and retries)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
mtagle: Yeah that's probably a good idea.

mtagle:
As discussed, I ran this against apilogger.log.api_log,apilogger.log.api_log_ro and everything seemed to work as expcted.  I'll also rerun after changes.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/partition/DynamicPartitioner.java
criccomini: Humm, reading more on:

https://cloud.google.com/bigquery/quota-policy

quotaExceeded seems to get triggered when > 100k writes/sec, or > 100mb/s to a single table. So you're right.. catching quotaExceeded should be done in BigQueryWriter. 

Based on this, I guess we'll only get 400's when the payload size is too large, so what you have sounds right.

criccomini: Ah.. What I was thinking of was this:

> HTTP request size limit: 10 MB. Exceeding this value will cause invalid errors.

And:

> invalid	400	This error returns when there is any kind of invalid input other than an invalid query, such as missing required fields or an invalid table schema. Invalid queries return an invalidQuery error instead.	

So not 403, "real" 400 errors that don't come from their front end LBs. They don't send these now because these requests don't make it to the BQ part of the code, but if they "fix" the front ends (increase the max size), then we'd start getting these, I think. So, I think we should try and catch them in the same spot as the "empty" 400 responses.

criccomini: (via https://cloud.google.com/bigquery/quota-policy and https://cloud.google.com/bigquery/troubleshooting-errors)

mtagle: We aren't handling rateLimitExceeded anywhere atm.  I think the correct place to handle that would be the BQWriter, but I don't think that change should be part of this RB.

If I'm interpreting correctly It sounds like you want to also catch 400s that are "invalid" (because that's what the spec says it will/should throw)?  I can do that.

criccomini: > We aren't handling rateLimitExceeded anywhere atm. I think the correct place to handle that would be the BQWriter, but I don't think that change should be part of this RB.

Agreed on both counts.

> If I'm interpreting correctly It sounds like you want to also catch 400s that are "invalid" (because that's what the spec says it will/should throw)? I can do that.

Correct.

mtagle: done

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/DynamicBatchWriter.java
```diff
@@ -0,0 +1,273 @@
...
+    if (exception.code() == BAD_REQUEST_CODE
+        && exception.error() == null
+        && exception.reason() == null) {
+      // 400 with no error or reason represents a request that is more than 10MB. This is not
+      // documented but is referenced slightly under "Error codes" here:
+      // https://cloud.google.com/bigquery/quota-policy
+      // (by decreasing the batch size we can eventually expect to end up with a request under
+      // 10MB)
+      return true;
+    } else if (exception.code() == BAD_REQUEST_CODE && INVALID_REASON.equals(exception.reason())) {
```
criccomini: Is "invalid" the reason or the error? Not sure how Google handles this.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -90,14 +92,11 @@
...
-  private static final ConfigDef.Validator MAX_WRITE_VALIDATOR =    ConfigDef.Range.atLeast(-1);
-  private static final ConfigDef.Importance MAX_WRITE_IMPORTANCE =  ConfigDef.Importance.MEDIUM;
-  private static final String MAX_WRITE_DOC =
-      "The maxiumum number of records to write at once to BigQuery, or -1 for no limit "
-          + "(cannot be zero)";
+  public static final String BATCH_WRITER_CONFIG = "batchWriter";
+  private static final ConfigDef.Type BATCH_WRITER_TYPE = ConfigDef.Type.CLASS;
+  public static final Class<? extends BatchWriter> BATCH_WRITER_DEFAULT = DynamicBatchWriter.class;
+  private static final ConfigDef.Importance BATCH_WRITER_IMPORTANCE = ConfigDef.Importance.HIGH;
+  private static final String BATCH_WRITER_DOC = "The batch writer class to be used.";
```
mtagle: Yep.  I'll make it more descriptive.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -135,31 +135,75 @@ private void updateAllPartitions(String topic, Map<TopicPartition, OffsetAndMeta
...
-      for (List<RowToInsert> partition : partitions) {
-        bigQueryWriter.writeRows(table, partition, topic, schemas);
-      }
+      batchWriter.writeAll(table, rows, topic, schemas);
       updateAllPartitions(tablesToTopics.get(table), offsets);
       return null;
     }
   }
 
+  private static class BatchWriterManager {
```
mtagle: If so, I think I can fix this just be having the map be a ConcurrentHashMap; it's the only mutable part of the class. Do you think that's a reasonable solution?

criccomini: I think you have to protect the "if not set, then set" logic as well

mtagle: Also, I think that each tableWriter is going to be writing to a different table, so there shouldn't actually be a whole lot of concerning overlap.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -135,31 +135,75 @@ private void updateAllPartitions(String topic, Map<TopicPartition, OffsetAndMeta
...
+    Map<TableId, BatchWriter<RowToInsert>> batchWriterMap;
+
+    public BatchWriterManager(BigQueryWriter bigQueryWriter,
+                              Class<BatchWriter<InsertAllRequest.RowToInsert>> batchWriterClass) {
+      this.bigQueryWriter = bigQueryWriter;
+      this.batchWriterConstructor = getBatchWriterConstructor(batchWriterClass);
+      batchWriterMap = new HashMap<>();
+    }
+
+    private static Constructor<BatchWriter<RowToInsert>>
```
mtagle: done

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/DynamicBatchWriter.java
mtagle: based on how it's documented I think it's the reason.  At least, quotaExceeded is a reason: https://github.com/wepay/kafka-connect-bigquery/blob/master/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/BigQueryWriter.java#L157

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
criccomini: Ya. I think the parallelism issue could still pop up with multiple partitions, though. If a single task is handling more than one partition for a given table (likely in our setup), there will be a race condition.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -192,7 +202,8 @@ public BatchWriterManager(BigQueryWriter bigQueryWriter,
     private void addNewBatchWriter(TableId tableId) {
       BatchWriter<RowToInsert> batchWriter;
       try {
-        batchWriter = batchWriterConstructor.newInstance(bigQueryWriter);
+        batchWriter = batchWriterConstructor.newInstance();
```
criccomini: I don't think you need batchWriterConstructor at all. You should be able to do `batchWriterClass.newInstance()` directly, and eliminate the whole getBatchWriterConstructor method.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkTaskConfig.java
```diff
@@ -120,12 +120,15 @@ private void checkAutoUpdateSchemas() {
   }
 
   private void checkMaxWrite() {
+    // noop (todo remove me)
```
mtagle: fixed!

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
mtagle: ah, I will fix this.

Issue #43, opened 2016-09-28
Regex table names


criccomini:
I'd like the ability to rename tables based on a regex in KCBQ.

We have topics in this format:

```
db.<connector>.<cluster>.<database>.<table>
```

That are getting fed data from Kafka connect sources. We're loading this data into BigQuery. When doing so, the dataset is:

```
db.<connector>.<cluster>.<database>
```

But the tables loaded inside the dataset are still:

```
db.<connector>.<cluster>.<database>.<table>
```

I'd prefer if they were just `<table>`. Being able to define a regex in config that maps topics in a dataset to table names would be helpful.

criccomini:
Closed by #44

PR #44, opened 2016-10-03
Add config mapping for topic to table names

willyhoang:
In cases where we do not want our table names to directly match our
topic names, a configuration should be available to rename the topics.


criccomini:
I think this is related to #43

criccomini:
@willyhoang Thanks for the PR. One observation is that this appears to be hard coded mappings. What do you think about supporting regex mappings, so you can do group-based mapping to bulk-rename tables that have a common pattern (as illustrated in #43)?

willyhoang:
Hey @criccomini, I didn't see that issue before but supporting regex mappings makes more sense to me! I'll try and make the necessary changes.

criccomini:
Thanks!

willyhoang:
Hey @criccomini could you help me by providing me with an example of a config value and how it would work? (maybe using the  "db.<connector>.<cluster>.<database>.<table>" from the issue")

criccomini:
Sure. One sec.

criccomini:
Suppose we have topics:

	db_debezium_identity_profiles_info
	db_debezium_identity_profiles_schools
	db_debezium_identity_profiles_addresses
	db_debezium_identity_profiles_jobs

And we have a dataset in BQ called:

	db_debezium_identity_profiles

We want tables in this dataset of:

	info
	schools
	addresses
	jobs

But right now the tables all have the same prefix as the dataset.

If we have a config like this:

	topicRename=db_debezium_identity_profiles_(.*)={0},...

Then it would give us the tables without their prefixes, which is what I'd want.

Couple of notes/thoughts:

1. The proposed config above does not encapsulate datasets. Topic rename for a given regex would span all datasets. There are a couple of ways that we could handle that, but it seemed to complicate things more than the benfit it provided. Open to suggests/discussion if anyone feels strongly about it, though.
2. I made up the {0} syntax for referencing regex groups. Java might have a standard for this, or we could use `String.format()` formatting, or some equivalen. Again, I don't feel too strongly on the style.

criccomini:
cc @mtagle ^^

criccomini:
@willyhoang LMK if you have any questions. :)

willyhoang:
@criccomini Definitely like the group based strategy more! Re Note 2. about syntax for regex groups it looks like $1, $2... is the standard (and easy to format using Matcher's replaceAll method).

criccomini:
Fantastic! Having a look now. :D

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -346,7 +414,14 @@ private String sanitizeTableName(String tableName) {
    *         name if sanitization is enabled.
    */
   public String getTableFromTopic(String topic) {
-    return getBoolean(SANITIZE_TOPICS_CONFIG) ? sanitizeTableName(topic) : topic;
+    String tableName = null;
+    Map<String, String> map = getTopicsToTables();
```
criccomini: Re-evaluating the map every time is inefficient (albeit not a huge cost). getTableFromTopic is called by getRecordTable, which is called on every single record that KCBQ writes. I think the map should be resolved once, and re-used.

criccomini: I kind of feel like moving this whole method to the BigQuerySinkTask would be best. Then you can resolve the map once as an instance variable, and then re-use it from within the class.

criccomini: Nit: `map` isn't a great variable name. how about `topicsToTablesMap` or some equivalent?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -346,7 +414,14 @@ private String sanitizeTableName(String tableName) {
...
+    tableName = map.get(topic);
+    if (tableName == null) {
```
criccomini: Seems like this table sanitization should be moved into getTopicsToTables or above. That way, it's done just once after the topic->table rewrite occurs.

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfigTest.java
```diff
@@ -102,6 +104,54 @@ public void testTablesToTopics() {
   }
 
   @Test
+  public void testTopicsToTables() {
```
criccomini: A test that validates that multiple groups in a single regex work would be good.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -346,7 +414,14 @@ private String sanitizeTableName(String tableName) {
...
+      tableName = getBoolean(SANITIZE_TOPICS_CONFIG) ? sanitizeTableName(topic) : topic;
+    }
+    return tableName;
+
```
criccomini: Nit: no extra space needed.

criccomini:
Also noticed that the tests are failing, which'll need to be addressed. :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
willyhoang: I tried moving the method over but realized it's also called in the BigQuerySinkConfig (getTablesToTopics) and in the BigQuerySinkConnector (ensureExistingTables). Since we want to be able to access the map in all three classes, is there a way to store the map as a variable in the config class?

criccomini: Let me have a look. I was trying to avoid storing any state in the config class since it feels wrong.

willyhoang: What if we had the getTablesFromTopic method take in a tablesToTopics map as a parameter. In the BigQuerySinkTask we can store an instance of that map (since like you said getTablesToTopics is called on every single record) and pass that instance into the method. 

In the BigQuerySinkConfig and in the BigQuerySinkConnector, we can just call the method like so: `getTableFromTopic(topic, getTopicsToTables())`, instantiating the map on the spot. In these two classes I'm assuming it's okay to do this since I think the method will only be called once in each.

criccomini: OK, I see the issue. My thought: move this stuff into BigQuerySinkConnector. The sink connector creates all of the tasks. I think the SinkTask should just take a static map from topic -> table. It shouldn't deal with doing any of the regex/resolving stuff. It should just do O(1) lookups on its map for each msg to get the table that a given msg is going into.

The SinkConnector can do all of the topic->table resolution one time, and then pass the final maps into the sink task. I *think* that should work, and will be a bit cleaner than dumping a bunch of utility logic into the config class (which is what we were doing even before you sent in this PR).

criccomini: OK, I'm resigned to this. Two options:

1. A util class (e.g. TopicToTableResolver) with a static method that takes a config and returns the map.
2. Cache the map in the config object

I'm partial to (1). The SinkConnector would use the util method to get the map for ensureExistingTables, and the SinkTask would use the util method to get the map and store it as an instance variable to be used in the class.

How's that sound?

criccomini: I suggest moving `getTablesToTopics` into SinkTask, since it's the only thing that uses it.

criccomini: (thanks for your patience :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -333,6 +388,19 @@ protected static void ensureValidMap(String name, List<String> values) {
     );
   }
 
+  /**
+   * Return a Map detailing which BigQuery table each topic should write to.
+   * @return A Map associating Kafka topic names to BigQuery table names.
+   */
+  public Map<String, String> getTopicsToTables() {
```
mtagle: instead of having two methods, onf ot which returns topics -> tables and the other which returns topics -> datasets, could we have a single method that returns topics to TableIds? (which includes dataset and table info)

criccomini: Yea that seems reasonable to me.

criccomini:
Awesome! This LGTM!

criccomini:
@willyhoang thanks so much for the PR!

criccomini:
Updated https://github.com/wepay/kafka-connect-bigquery/wiki/Connector-Configuration with the new config option! :)

willyhoang:
Awesome!

criccomini:
@mtagle can you publish an 0.4.0 release to Sonatype?

PR #48, opened 2016-10-14
Partition BQ Tables

mtagle:
Partition tables.
Partition rows in these tables by their insertion times.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableResolver.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableResolver.java
@@ -96,7 +96,7 @@ public class TopicToTableResolver {
     int month = localDate.getMonthValue();
     int day = localDate.getDayOfMonth();
     sb.append(year);
-    sb.append(month + 1); // java 0 indexes months; google 1 indexes months.
+    sb.append(month);
     sb.append(day);
     String partitionedTableName = sb.toString();
     if (baseTableId.project() == null) {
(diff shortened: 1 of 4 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableResolver.java
```diff
@@ -83,6 +87,38 @@
     return matches;
   }
 
+  // package private for testing
+  static TableId getPartitionedTableName(TableId baseTableId, Date date) {
+    StringBuilder sb = new StringBuilder();
+    String baseTableName = baseTableId.table();
+    sb.append(baseTableName);
+    sb.append("$");
+    Calendar cal = Calendar.getInstance();
```
criccomini: I'm nervous that this code is going to use timezones, and we want everything to be UTC. Can you force it to use UTC?

criccomini: Should also move away from `java.util.date` to new Java 8 `java.time` classes.

criccomini:
Also, can you confirm that we don't need to use (and don't get any advantage) from [template tables](https://cloud.google.com/bigquery/streaming-data-into-bigquery#template-tables)? My reading is that we don't need to use that feature, but want to make sure.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableResolver.java
mtagle: fixed

mtagle:
As a test I just tried inserting into a partitioned table with a template suffix ($20161014) and it didn't work, because table names can't contain special characters like $.  That section doesn't mention anything about partitioned tables, so I'm going to assume that this isn't meant to be used with partitioned tables.

criccomini:
LGTM go ahead and merge in after you've tested locally against a real GCP project. Also, the integration tests should validate as well.

PR #50, opened 2016-10-20
Replace Buffers with simplier lists

mtagle:
These lists function similarly to buffers but do not error if they are added to when they have excess.


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -317,8 +323,8 @@ public void put(Collection<SinkRecord> records) {
         tableRows.add(getRecordRow(record));
       }
 
-      buffer.buffer(tableRows);
-      if (buffer.hasExcess()) {
+      buffer.addAll(tableRows);
+      if (bufferFull(buffer.size())) {
         pauseAllPartitions(tableIdsToTopics.get(table));
```
criccomini: Can you add a log.info() in `pauseAllPartitions` that says which partitions are being paused?

PR #51, opened 2016-10-24
Fix dynamic batch writer

mtagle:
Introduce a PartitionedTableId.  Like a TableId, but it stores the partition information of the table separately from the table.
Make DynamicBatchWriter work better with partitioned tables.
Remove potential for memory leak in how DynamicBatchWriters for tables are saved.
All writes to the same base table, irrespective of the partition, will use the same DynamicBatchWriter.

Add tests.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -155,12 +155,12 @@ public class BigQuerySinkTask extends SinkTask {
   }
 
   /**
-   * A class for keeping track of the BatchWriters for each partitionedTableId.
+   * A class for keeping track of the BatchWriters for each base table.
    */
   private static class BatchWriterManager {
     private final BigQueryWriter bigQueryWriter;
     private final Class<BatchWriter<InsertAllRequest.RowToInsert>> batchWriterClass;
-    // map from base partitionedTableId id to the batchWriter for that base partitionedTableId
+    // map from base TableId to the batchWriter for that base table
     private Map<TableId, BatchWriter<RowToInsert>> baseTableBatchWriterMap;
 
     /**
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -34,6 +34,7 @@ import com.wepay.kafka.connect.bigquery.exception.BigQueryConnectException;
 import com.wepay.kafka.connect.bigquery.exception.SinkConfigConnectException;
 
 import com.wepay.kafka.connect.bigquery.utils.MetricsConstants;
+import com.wepay.kafka.connect.bigquery.utils.PartitionedTableId;
 import com.wepay.kafka.connect.bigquery.utils.TopicToTableResolver;
 import com.wepay.kafka.connect.bigquery.utils.Version;
 
@@ -91,10 +92,10 @@ public class BigQuerySinkTask extends SinkTask {
   private final BigQuery testBigQuery;
   private BigQuerySinkTaskConfig config;
   private RecordConverter<Map<String, Object>> recordConverter;
-  private Map<TableId, List<RowToInsert>> tableBuffers;
+  private Map<PartitionedTableId, List<RowToInsert>> tableBuffers;
   private Map<TableId, Set<Schema>> tableSchemas;
   private BatchWriterManager batchWriterManager;
-  private Map<TableId, String> tableIdsToTopics;
+  private Map<TableId, String> baseTableIdsToTopics;
   private Map<String, TableId> topicsToBaseTableIds;
   private Metrics metrics;
   private Sensor rowsRead;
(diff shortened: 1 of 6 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -125,41 +126,42 @@ private void updateAllPartitions(String topic, Map<TopicPartition, OffsetAndMeta
...
-      updateAllPartitions(tableIdsToTopics.get(table), offsets);
+      batchWriter.writeAll(partitionedTableId.getFullTableId(), rows, topic, schemas);
+      updateAllPartitions(baseTableIdsToTopics.get(partitionedTableId.getBaseTableId()), offsets);
       return null;
     }
   }
 
   /**
-   * A class for keeping track of the BatchWriters for each table.
+   * A class for keeping track of the BatchWriters for each partitionedTableId.
```
criccomini: I don't actually see any usage of partitionedTableId in this class. Is this comment accurate?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -207,23 +209,23 @@ private void resumeAllPartitions() {
...
                 offsets,
-                tableIdsToTopics.get(table),
-                tableSchemas.get(table)
+                baseTableIdsToTopics.get(table.getBaseTableId()),
+                tableSchemas.get(table.getBaseTableId())
             )
         );
         // wipe buffer
-        tableBuffers.put(table, getNewBuffer());
+        tableBuffers.remove(table);
```
criccomini: Curious about put -> remove

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/PartitionedTableId.java
```diff
@@ -0,0 +1,225 @@
...
+
+import com.google.cloud.bigquery.TableId;
+
+import java.time.Clock;
+import java.time.LocalDate;
+
+/**
+ * A TableId with separate base table name and partition information.
+ */
+public class PartitionedTableId { // todo make tests for me
```
criccomini: TODO is out dated? I see tests below. :)

mtagle: oops, yes.  will remove.

criccomini:
LGTM! If you've tested locally, then go ahead and merge.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -233,18 +235,23 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
...
       }
     }
     resumeAllPartitions();
   }
 
-  private TableId getRecordTable(SinkRecord record) {
-    return TopicToTableResolver.getPartitionedTableName(topicsToTableIds.get(record.topic()));
+  private PartitionedTableId getRecordTable(SinkRecord record) {
+    TableId baseTableId = topicsToBaseTableIds.get(record.topic());
+    return new PartitionedTableId.Builder(baseTableId).setDayPartitionForNow().build();
```
cleverscale: Regarding `setDayPartitionForNow()`, should it consider the record timestamp instead ? It seems that if I consume old records of a topic (or use a custom timestamp extractor), the data will always be inserted in the partition of the current day.

nfo: Regarding `setDayPartitionForNow()`, should it consider the record timestamp instead ? It seems that records are always inserted in the partition for the current day; But I could consume a topic with "old" records, or use a custom timestamp extractor.
@mtagle @criccomini

criccomini:
@nfo I think there are arguments for both use cases.

The reason that we opted for wall-clock time has mostly to do with the data that we're loading. We're loading database changelog data, which contains inserts, updates, and deletes. We then periodically want to go and clean out all of the older mutations for each primary key.
For example, if we had a row with a MySQL primary key of 1234, and it had:

insert@time t1
update@time t2
update@time t3

We really only need the last update@time t3 in order to display the current snapshot of the table, so we periodically want to go and clean out the older rows for cost and performance reasons.  Since BigQuery is basically append-only (i.e. it's pretty difficult to "delete" rows), having everything appended to the end of the log means that it's much easier for us to go and drop older partitions (since they'll never be written to again after the day rolls over). It also makes it easier for us to overwrite data (since we just have to append it to the current day).

In a sense, we treat our BigQuery tables exactly like Kafka log-compacted topics. We always append to the head of the table, and we do compaction and deletion operations on the older partitions.

That said, I recognize that this is not a common way of dealing with data. I also agree that it makes a lot of sense to want to use event time, rather than wall clock time for partitioning. I wouldn't be opposed to a patch that allows for both use cases. That said, I do think we should continue defaulting to the current behavior.

As for implementation, I can think of at least two ways to pursue this off the top of my head:

1. You could add a pluggable partition assignment interface, and then individuals could implement their own class to suck the timestamp out of the event and return it. This is a bit clunky, though, because presumably you'd have to implement one class per schema type (since topics might have different schemas with different timestamp fields). You'd then have to assign each class to a topic (or group of topics) via the config.
2. You could take a more rigid approach where users could configure the timestamp path for the SinkRecord in a topic via the configuration. Rather than specifying a custom class that does the partitioning (as in (1), above), they'd just specify topic (or topic-regex) to timestamp paths in the configuration. For example, something like `timestamp.extractors=my.topic.regex=producerHeader.timestamp`. One thing you'd have to cope with here is the fact that you don't know what data type the timestamp is (Date, Timestamp, long utc seconds, long utc milliseconds, iso string, etc).

Some food for thought.

PR #66, opened 2016-11-04
Updating Schemas for partitioned tables should no longer fail.

mtagle:
primarily pushing down switching from partitionedTableId to TableId further down the stack.


criccomini:
LGTM

PR #67, opened 2016-11-07
I-62: More logging when pausing partitions

mtagle:
Log how long a partition has been running when it has been paused, how long it has been paused when it is resumed, and the buffer size when it is paused.

The basic code functionality seems untouched (at least I was able to run kcbq locally without issue), but my own local changes don't seem to be triggering pauses, so I'm less able to see if the logging messages are correct.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -442,14 +442,8 @@ public class BigQuerySinkTask extends SinkTask {
 
   private class TopicPartitionManager {
 
-    // keeping this roughly as is is almost certainly fine though.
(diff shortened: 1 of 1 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -444,4 +434,81 @@ public String version() {
...
+    private Map<TopicPartition, State> topicStates;
+    private Map<TopicPartition, Long> topicChangeNano;
+
+    public TopicPartitionManager() {
+      topicStates = new HashMap<>();
+      topicChangeNano = new HashMap<>();
+    }
+
+    public void pause(String topic, int topicBufferSize) {
+      Long now = System.nanoTime();
```
criccomini: Curious why nano vs System.currentTimeMillis()

PR #72, opened 2016-11-17
Remove the retrieveSchema method that takes in the set of failed Sche…

mtagle:
…mas, since it's just ignored anyway.

Less stuff to pass around.

### DIFF
--- a/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
+++ b/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
@@ -20,18 +20,7 @@ public interface SchemaRetriever {
   public void configure(Map<String, String> properties);
 
   /**
-   * Retrieve the most current schema for the given topic, in order to perform an update in
-   * BigQuery.
-   * @param table The table that will be updated.
-   * @param topic The topic to retrieve a schema for.
-   * @param schemas The set of unique Schemas for the data that caused the insert failure.
-   * @return The Schema for the given table.
-   */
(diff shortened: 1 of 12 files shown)


criccomini:
I believe @C0urante and I included the retrieveSchema method with the schema param to allow for cases where there is no schema registry (e.g. if you're using a protobuf or thrift serialization). In such cases, it would be useful to know the schema that failed.

That said, all of this is conjecture, since we only support Avro right now. I'm curious what motivated the deletion?

mtagle:
The schemas are only "used" in on specific circumstance: When you fail to write to a table because of invalid schema errors, and you have to update the BQ table.  The perform writeReqeust method in BQ writer takes in schemas but it only the AdaptiveBQWriter actually "uses" it.  The schemas are not needed to actually write data to bq in 99.9+% of cases and yet we are carting them around everywhere nonetheless. 

"use" of course, is in parens because  it's not actually used in any circumstances at the moment.

I'm still not totally clear on when having the set of schemas that failed would be particularly useful. The interface already requires that any implementation would have to implement a method that allows you to get a schemas for writing a new table, when you don't have access to any "failed" schemas.

mtagle:
A larger objection I have to carting around the schemas (and also the topic) is that it is very unclear why they are needed to perform a write request. I don't think this would be limited to just me either, because they aren't really needed except in 1 small case. Even in that case, they are just used as keys to get some other information that's more obviously relevant to writing data to BQ.  If it was reasonable to pre-emptively get the schema that matches the information we are trying to write that would be more acceptable to me, because the schema of the data we are trying to write is more obviously relevant to writing data than the topic of the original source of the data, or the set of all schemas we have, in the past, used to write the data.

criccomini:
I'm convinced. :) +1 Go ahead and merge!

PR #73, opened 2016-12-03
I-63: Trigger flushes on put

mtagle:
put now places tasks into a task queue that is executed by a threadpoolexecutor.

flush makes sure everything in that queue is finished before returning.

I intend to add some more stuff to this, and test it locally, but this is the expected structure of the code.  It is mostly complete, but I want to add some functionality to the TableWriter to be able to split up requests that fail due to being too large (similarly to what DynamicBatchWriter did).  Right now if put has too large of a request for a single table it just won't be able to continue.


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkTaskConfig.java
```diff
@@ -40,14 +40,27 @@
...
+      ConfigDef.Importance.MEDIUM;
+  private static final String THREAD_POOL_SIZE_DOC =
+      "The size of the BigQuery write thread pool. This establishes the maximum number of "
+      + "concurrent writes to BigQuery.";
+
+  public static final String TOPIC_MAX_THREADS_CONFIG =               "tableMaxThreads";
+  private static final ConfigDef.Type TOPIC_MAX_THREADS_TYPE =        ConfigDef.Type.INT;
+  public static final Integer TOPIC_MAX_THREADS_DEFAULT =             7;
+  private static final ConfigDef.Validator TOPIC_MAX_THREADS_VALIDATOR =
+      ConfigDef.Range.atLeast(1); // also no more than THREAD_POOL_SIZE?
```
criccomini: I think there's a between method:

https://kafka.apache.org/0100/javadoc/org/apache/kafka/common/config/ConfigDef.Range.html#between(java.lang.Number,%20java.lang.Number)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/KCBQThreadPoolExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+  protected void beforeExecute(Thread thread, Runnable runnable) {
+    super.beforeExecute(thread, runnable);
+
+    if (runnable instanceof TableWriter) {
+      TableWriter tableWriter = (TableWriter) runnable;
+      String topic = tableWriter.getTopic();
+      activeThreadTopicCount.putIfAbsent(topic, new AtomicInteger(0));
+
+      int topicThreadCount = activeThreadTopicCount.get(topic).incrementAndGet();
+      if (topicThreadCount > topicThreadLimit) {
```
criccomini: Is there an off by one here? At this point, if topic thread limit is 3, and the incrementAndGet returns 3, I don't think that this class will pause the topic. If it doesn't, then the next time around incrementAndGet can return 4, which would mean we'd be running one extra thread.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/KCBQThreadPoolExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+
+  /**
+   * Wait for all the currently queued tasks to complete, and then return.
+   *
+   * @throws BigQueryConnectException if any of the tasks failed.
+   * @throws InterruptedException if interrupted while waiting.
+   */
+  public void awaitCurrentTasks() throws InterruptedException, BigQueryConnectException {
+    int maximumPoolSize = getMaximumPoolSize();
+    CountDownLatch countDownLatch = new CountDownLatch(maximumPoolSize);
```
criccomini: I don't quite follow this code. I think the trick that you're playing is that you're stuffing the max pull size number of countdown latches into the thread pool and awaiting on it to make sure that everything prior to it has finished.

What was the reason for doing this vs. calling join() on each remaining thread in the pool directly?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/KCBQThreadPoolExecutor.java
```diff
@@ -0,0 +1,218 @@
...
+                        .collect(Collectors.toList()));
+    return String.join(", ", exceptionTypeStrings);
+  }
+
+  private enum State {
+    PAUSED,
+    RUNNING
+  }
+
+  private static class TopicPartitionManager {
```
criccomini: Is this going to be removed from BigQuerySinkTask?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/KCBQThreadPoolExecutor.java
mtagle: I'm not familiar with join, so let me know if I'm totally wrong in any assumptions I'm making here.  It looks like it's a method on Thread that just waits for a thread to die.  Since I'm using an executor, I don't actually have access to handles to the threads, so I can't call .join() on all the threads.  Even if I did, the amount of time that would have to pass before we would know that a thread was "done" would depend on the configured keepalive time of threads in the pool, which I think would be not great.

There are other ways to do this with an executor, I guess, like looping while constantly checking if the executor is no longer executing, but this seemed like the best solution.

mtagle:
commit message here is just that the tests work, but local integration tests (with multiple tasks and threads) also worked.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/KCBQThreadPoolExecutor.java
criccomini: > I don't actually have access to handles to the threads

I think you can get them in beforeExecute.

It just bugs me that you had to write this countdown thingy to do something that seems like it should support out of the box. I Googled around a bit and found [this](http://stackoverflow.com/questions/3929361/how-to-wait-for-all-tasks-in-an-threadpoolexecutor-to-finish-without-shutting-do) discussion, which uses a similar strategy, albeit with futures. Grabbing the threads, and calling join on all of them would be roughly equivalent.

Anyway, if there's a way to eliminate the need for `CountDownRunnable`, I think it'd be worth pursuing.

criccomini: Another option would be to just .shutdown() the executor in `flush()`, then `awaitTermination`, then create a new one. This seems kind of appealing because it's so simple. :)

mtagle: Both join methods (by whatever means we would get the thread handles) and shutting down the executor rely on threads dying, which is not great because then we'll have to recreate them all when we end up putting again. (which is expensive)

The countdown latch makes sure that no threads need to die (and then be re-created expensively soon after).

if the concern is that the countdown is too confusing, I can add some comments around it.

criccomini:
How long does it take to create the thread executor again? With default flush interval, this would happen once every 30s.

criccomini:
Was thinking more about the countdown latch strategy. I don't think it'll work. If you have a pool of 3 threads, and you have all three occupied with work:

q=[]
t0 = w0
t1 = w1
t2 = w2

And you add in your three count down latch threads to the queue:

q=[cd0, cd1, cd2]
t0 = w0
t1 = w1
t2 = w2

Then w0 and w1 finish:

q=[cd0, cd1, cd2]
t0 = empty
t1 = empty
t2 = w2

The executor will assign cd0 and cd1:

q=[cd2]
t0 = cd0
t1 = cd1
t2 = w2

Then the cd's finish:

q=[cd2]
t0 = empty
t1 = empty
t2 = w2

And cd2 runs:

q=[]
t0 = empty
t1 = cd2
t2 = w2

And when cd2 finishes, flush will return, even though w2 is still running. This is based on my understanding of the design, so it might not be accurate, but as I understand it, it seems like it won't work.

mtagle:
The count down latches don't "finish" till all other latches have counted down.

mtagle:
It doesn't take long to create an executor, but it does take relatively long to spin up a new thread; this is the entire purpose behind the existance of a threadpool.

criccomini:
> It doesn't take long to create an executor, but it does take relatively long to spin up a new thread; this is the entire reason behind the existence of threadpools.

Sure, but we're still talking milliseconds right?

http://stackoverflow.com/questions/5483047/why-is-creating-a-thread-said-to-be-expensive

> A Google search found me an old benchmark that reports a thread creation rate of ~4000 per second on a Sun Java 1.4.1 on a 2002 vintage dual processor Xeon running 2002 vintage Linux.

criccomini:
> The count down latches don't "finish" till all other latches have counted down.

I don't follow this, either.  AFAICT, the latches are counted down by the CountDownRunnable class. These runnables can all run and count down provided that just a single thread is free in the thread pool, right? There's no affinity between the runnable and the threads. My example above illustrates the issue I'm talking about.

mtagle:
initial benchmarking stats look good. close to 3m rows in 15 minutes.

I should make the logging a bit less verbose though: 

```
...
[2016-12-08 16:27:06,534] INFO Wrote 572 rows over 1 successful calls and 0 failed calls. (com.wepay.kafka.connect.bigquery.write.batch.TableWriter:91)
[2016-12-08 16:27:07,068] INFO Wrote 304 rows over 1 successful calls and 0 failed calls. (com.wepay.kafka.connect.bigquery.write.batch.TableWriter:91)
[2016-12-08 16:27:07,156] INFO Wrote 360 rows over 1 successful calls and 0 failed calls. (com.wepay.kafka.connect.bigquery.write.batch.TableWriter:91)
...
```

Probably turn this log message to debug and make failure rate a metric.

criccomini:
re: countdown latch working, I see now that the latches all .await before they release the thread, which forces a full flush. :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -112,124 +96,26 @@ public BigQuerySinkTask() {
...
-      batchWriter.writeAll(partitionedTableId, rows, topic);
-      updateAllPartitions(baseTableIdsToTopics.get(partitionedTableId.getBaseTableId()), offsets);
-      return null;
-    }
+    topicPartitionManager.resumeAll();
   }
 
   /**
-   * A class for keeping track of the BatchWriters for each base table.
+   * This really doesn't do much and I'm not totally clear on whether or not I need it.
```
ewencp: Very late to the game here, but since I found this via https://issues.apache.org/jira/browse/KAFKA-6148 figured I'd mention that this code should not, in fact, be needed. If you actually flush all data in `flush()` and do not throw an exception, the correct offsets will be automatically committed. You only need to use `context.offset` if you want to fully manage offsets yourself.

C0urante: I could be off-base here but I think the reason we initially included manual updating of offsets was to help push things forward in the event of a failed `flush()` due to an excessive volume of data. If too many records were meant to be written in a call to `flush()` and an exception ended up being thrown, the manual updates would prevent the next call to `flush()` from doing the exact same thing and trapping the sink task in an endless loop. If there are already steps in place for the Connect framework to handle this case automatically then the code block is probably obsolete and can be removed safely.

ewencp: If an exception is thrown from flush, the connector will still continue from where it was, it would just be unable to commit offsets. You can see how that is handled [here](https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L343-L359) though now `preCommit()` wraps `flush()` by default, so that's the relevant call. Note that we don't modify the consumer in the event of failure, we just return from the call, so the consumer continues where it left off. Perhaps there are some javadoc updates in the SinkTask interface that would help clarify this.

For this connector there's probably a better fix [KIP-89](https://cwiki.apache.org/confluence/display/KAFKA/KIP-89%3A+Allow+sink+connectors+to+decouple+flush+and+offset+commit) allows the sink task better control over how offset commits work, allowing it to, e.g., not immediately flush data but just return what safe offsets to commit are. You have to be a bit careful to not allow your backlog to grow indefinitely (i.e. don't consistently accept data from Kafka faster than you can write it to your sink), but this allows you to just very quickly return from `preCommit` and never have to fully flush the pipeline.

In any case, we tracked down the issue to actually be that the build process for this connector is incorrectly including `connect-api` instead of marking it provided, which in turn pulls in a copy of `kafka-clients` which doesn't play well with the new classloader isolation. We're going to make a fix to Connect so that even if you include the jar we'll load the clients classes from the right place, but Connectors shouldn't be including a copy of either jar since that can make them only work with the specific version of Connect they were compiled with when using loading of plugins via CLASSPATH.

I filed https://github.com/wepay/kafka-connect-bigquery/pull/100 which should fix the issue except that debezium has the same problem. I don't know gradle well enough to deal with excluding all transitive dependencies to make this easily fixed without also fixing the debezium issue.

PR #80, opened 2017-02-03
Fix two issues around updating schemas:

mtagle:
1. Ignore error message "stopped" if we have previously seen schema invalid errors while checking if an error message is only about an invalid schema.
2. Set a limit on retrying row insertions after attempting to update the schema of a BQ table.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
@@ -98,7 +98,7 @@ public class AdaptiveBigQueryWriter extends BigQueryWriter {
         throw new BigQueryConnectException(writeResponse.insertErrors());
       }
       attemptCount++;
-      if (attemptCount > AFTER_UPDATE_RETY_LIMIT) {
+      if (attemptCount >= AFTER_UPDATE_RETY_LIMIT) {
         throw new BigQueryConnectException("Failed to write rows after BQ schema update within "
                                            + AFTER_UPDATE_RETY_LIMIT + " attempts.");
       }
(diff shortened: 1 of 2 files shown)


mtagle:
In my testing, I only saw invalid schemas followed by stopped.  I'm not sure if you can have stopped before invalid schemas, but I wanted to keep the checking as strict as possible.

I didn't really find any official documentation on this (this is all I found: http://stackoverflow.com/questions/21461317/bigquery-data-errorproto-reason-stopped), so I'm open to opinions.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -76,10 +80,15 @@ public void performWriteRequest(PartitionedTableId tableId,
     // BigQuery schema updates taking up to two minutes to take effect
     if (writeResponse.hasErrors()
         && onlyContainsInvalidSchemaErrors(writeResponse.insertErrors())) {
-      schemaManager.updateSchema(tableId.getBaseTableId(), topic);
+      try {
+        schemaManager.updateSchema(tableId.getBaseTableId(), topic);
```
criccomini: Curious about this change. Is BigQueryException a checked exception?

mtagle: Previously, this error would percolate up and be caught by the batch size retry logic, which looks for a 400 with an "invalid" message.  If you try to update a table with an invalid schema, BigQuery will return a 400 "invalid". So, without this, there is some weirdness and a lack of clarity whenever this error pops up. This prevents that.

criccomini:
LGTM go ahead and merge

PR #81, opened 2017-02-13
Support array of structs in conversion

willyhoang:
The record converter code previously assumed that an array value would
only be an array of primitives. This is not true as there can be Kafka
records with arrays of records. In this case, we should call the convert
method on every object in the array.

Resolves https://github.com/wepay/kafka-connect-bigquery/issues/82.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java
@@ -88,16 +88,16 @@ public class BigQueryRecordConverter implements RecordConverter<Map<String, Obje
     }
     Schema.Type kafkaConnectSchemaType = kafkaConnectSchema.type();
     switch (kafkaConnectSchemaType) {
-      case STRUCT:
-        return convertStruct(kafkaConnectObject, kafkaConnectSchema);
+      case ARRAY:
+        return convertArray(kafkaConnectObject, kafkaConnectSchema);
       case MAP:
         return convertMap(kafkaConnectObject, kafkaConnectSchema);
+      case STRUCT:
+        return convertStruct(kafkaConnectObject, kafkaConnectSchema);
       case BYTES:
         ByteBuffer byteBuffer = (ByteBuffer) kafkaConnectObject;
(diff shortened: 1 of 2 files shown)


criccomini:
Thanks for posting this! We are actually observing a problem in this area in our production environment right now. :) I think @mtagle is having a look at the PR.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java
```diff
@@ -97,7 +97,7 @@ private Object convertObject(Object kafkaConnectObject, Schema kafkaConnectSchem
         byte[] bytes = byteBuffer.array();
         return Base64.getEncoder().encodeToString(bytes);
       case ARRAY:
```
mtagle: lets rearrange these so the order is Array, Map, Struct, Bytes, Boolean ... so the organization is a little better.

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverterTest.java
```diff
@@ -336,6 +336,53 @@ public void testIntegerArray() {
   }
 
   @Test
+  public void testStructArray() {
+    final String innerFieldStringName = "InnerString";
+    final String innerFieldIntegerName = "InnerInt";
+    final String innerStringValue = "42";
```
mtagle: just for clarity's sake, could we make this more obviously a string?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java
willyhoang: Sounds good to me!

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverterTest.java
willyhoang: Sure thing.

PR #84, opened 2017-04-03
Upgrade confluent version to 3.2

mtagle:


C0urante:
Has anybody checked to see if the cherry-picking outlined in the README for Schema Registry is still necessary to build the connector? Might want to nix that section now if the bump to 3.2 takes care of it.

criccomini:
LGTM! Go ahead and merge

criccomini:
@C0urante good question... I don't think it is, but should double check. If not, @mtagle can you do a separate PR for that?

mtagle:
@criccomini will do.

PR #89, opened 2017-06-15
upgrade kafka-client

RaiseJohnBarry:
connect-distributed does not work in current configuration; needs a
10.2.1 api.


criccomini:
Note: we are running KCBQ in distributed mode right now. Our KC is on 0.10.2.0, though.

PR #99, opened 2017-10-27
Fix bug where Debezium Logical Converters were not used correctly.

mtagle:
Also fix some found issues with the debezium date type converter.

Related tests.


criccomini:
LGTM!

PR #100, opened 2017-11-04
Fix dependencies so connect-api is a provided dependency.

ewencp:
https://issues.apache.org/jira/browse/KAFKA-6148 is an example where using plugin.path for loading can cause conflicts because connect-api pulls in kafka-clients and a single class was missing special handling to force it to be loaded by the system classloader.

However, connect-api shouldn't be included in packaged versions of connectors anyway, even with CLASSPATH loading, because that could cause potential conflicts between the framework's version and the version the plugin is compiled against. Connect takes great pain to remain backwards compatible so connectors can continue to work with newer versions of the Connect framework without recompiling / changing dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -186,8 +186,7 @@ project(':kcbq-connector') {
 
         compile group: 'io.debezium', name: 'debezium-core',  version:'0.4.0'
 
-        compile group: 'org.apache.kafka', name: 'connect-api', version: '0.10.2.1'
-        compile group: 'org.apache.kafka', name: 'kafka-clients', version: '0.10.2.1'
+        compileOnly group: 'org.apache.kafka', name: 'connect-api', version: '0.10.2.1'
 
         compile group: 'org.slf4j', name: 'slf4j-api', version: '1.6.1'
         compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.6.1'
@@ -234,7 +233,7 @@ project('kcbq-api') {
     dependencies {
         compile group: 'com.google.cloud', name: 'google-cloud', version: '0.10.0-alpha'
 
-        compile group: 'org.apache.kafka', name: 'connect-api', version: '0.10.2.1'
+        compileOnly group: 'org.apache.kafka', name: 'connect-api', version: '0.10.2.1'
     }
 
     artifacts {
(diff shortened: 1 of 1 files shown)


ewencp:
Note that while this is the correct fix for this repo, the debezium dependency [has the same problem](https://issues.jboss.org/browse/DBZ-285) so that dependency would also need to be upgraded and this patch adjusted before it would make sense to merge. As it stands, this patch gets rid of the version pulled in by this connector, but the debezium transitive dependencies end up pulling in a different version.

build.gradle
```diff
@@ -186,8 +186,7 @@ project(':kcbq-connector') {
 
         compile group: 'io.debezium', name: 'debezium-core',  version:'0.4.0'
 
-        compile group: 'org.apache.kafka', name: 'connect-api', version: '0.10.2.1'
-        compile group: 'org.apache.kafka', name: 'kafka-clients', version: '0.10.2.1'
```
mtagle: it looks to me like we use this in BigQuerySinkTask? https://github.com/wepay/kafka-connect-bigquery/blob/master/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java#L44

ewencp: All Kafka dependencies get pulled in transitively as needed. The `SinkTask` use is via an interface included in the Connect APIs, so connectors should not be explicit in the version they depend on.

I regret letting this through as at the time I think we didn't realize we were exposing the transitive dependency on Kafka clients (and the dependence is *very* minimal). But for connectors, they should rely on the transitive dependency. It is already always guaranteed to be pinned to the same version given that the two projects are maintained in the same repo.

mtagle: but should kafka-clients be pulled in as compileOnly, like connect-api, or completely removed like it is now?

It also looks like the DBZ issue you linked that would have to be resolved first is now resolved. So, is this patch currently okay to merge or is there anything else on the dbz side that needs to be done beforehand?

RaiseJohnBarry:
I'm fairly certain this won't alleviate the problem. 
We've replaced the updateOffsets() call with super.flush() to fix it. (in BigQuerySinkTask)
(I'd PR but we're still testing it with large topics/tables)

PR #101, opened 2017-11-06
Implement in-memory schema retriever

makearl:
This change implements an in-memory schema retriever which caches the last seen schema for a given topic (inspiration drawn from how the JDBC sink connector handles schema updates).

Our use case for this schema retriever is to be able to support creating and updating table schemas for serialization formats that do not use the schema registry.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -69,6 +69,7 @@ public class BigQuerySinkTask extends SinkTask {
   private static final Logger logger = LoggerFactory.getLogger(BigQuerySinkTask.class);
 
   private final BigQuery testBigQuery;
+  private SchemaRetriever schemaRetriever;
   private BigQueryWriter bigQueryWriter;
   private BigQuerySinkTaskConfig config;
   private RecordConverter<Map<String, Object>> recordConverter;
@@ -81,11 +82,13 @@ public class BigQuerySinkTask extends SinkTask {
 
   public BigQuerySinkTask() {
     testBigQuery = null;
+    schemaRetriever = null;
   }
 
   // For testing purposes only; will never be called by the Kafka Connect framework
-  BigQuerySinkTask(BigQuery testBigQuery) {
+  BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever) {
     this.testBigQuery = testBigQuery;
+    this.schemaRetriever = schemaRetriever;
   }
 
   @Override
@@ -150,6 +153,10 @@ public class BigQuerySinkTask extends SinkTask {
     for (SinkRecord record : records) {
       if (record.value() != null) {
         PartitionedTableId table = getRecordTable(record);
+        if (schemaRetriever != null) {
+          schemaRetriever.setLastSeenSchema(table.getBaseTableId(), record.topic(), record.valueSchema());
+        }
+
         if (!tableWriterBuilders.containsKey(table)) {
           TableWriter.Builder tableWriterBuilder =
               new TableWriter.Builder(bigQueryWriter, table, record.topic());
@@ -191,7 +198,7 @@ public class BigQuerySinkTask extends SinkTask {
   }
 
   private SchemaManager getSchemaManager(BigQuery bigQuery) {
-    SchemaRetriever schemaRetriever = config.getSchemaRetriever();
+    schemaRetriever = config.getSchemaRetriever();
     SchemaConverter<com.google.cloud.bigquery.Schema> schemaConverter =
         config.getSchemaConverter();
     return new SchemaManager(schemaRetriever, schemaConverter, bigQuery);
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/retrieve/MemorySchemaRetriever.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/retrieve/MemorySchemaRetriever.java
@@ -1,0 +1,58 @@
+package com.wepay.kafka.connect.bigquery.retrieve;
+
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.TableId;
+import com.wepay.kafka.connect.bigquery.api.SchemaRetriever;
+import org.apache.kafka.common.cache.Cache;
+import org.apache.kafka.common.cache.LRUCache;
+import org.apache.kafka.common.cache.SynchronizedCache;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Map;
+
+/**
+ * Uses the Confluent Schema Registry to fetch the latest schema for a given topic.
+ */
+public class MemorySchemaRetriever implements SchemaRetriever {
+  private static final Logger logger = LoggerFactory.getLogger(MemorySchemaRetriever.class);
+  private static final int CACHE_SIZE = 1000;
+  private Cache<String, Schema> schemaCache;
+
+  /**
+   * Only here because the package-private constructor (which is only used in testing) would
+   * otherwise cover up the no-args constructor.
+   */
+  public MemorySchemaRetriever() {
+  }
+
+  private String getCacheKey(String tableName, String topic) {
+    return new StringBuilder(tableName).append(topic).toString();
+  }
+
+  @Override
+  public void configure(Map<String, String> properties) {
+    schemaCache = new SynchronizedCache<>(new LRUCache<String, Schema>(CACHE_SIZE));
+  }
+
+  @Override
+  public Schema retrieveSchema(TableId table, String topic) {
+    String tableName = table.getTable();
+    Schema schema = schemaCache.get(getCacheKey(tableName, topic));
+    if (schema != null) {
+      logger.info("Using cached schema");
+      return schema;
+    }
+
+    logger.info("Using empty struct");
+    return SchemaBuilder.struct().build();
+  }
+
+  @Override
+  public void setLastSeenSchema(TableId table, String topic, Schema schema) {
+    logger.info("Updating last seen schema to " + schema.toString());
+    schemaCache.put(getCacheKey(table.getTable(), topic), schema);
+  }
+}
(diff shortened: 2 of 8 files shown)


mtagle:
This sounds great, but I'm a little confused at a high-level.

Can you explain how the memory retriever will return the most recent schema, and how we can be sure the cache isn't stale?

makearl:
For every incoming message, we insert the schema for the message for that table and topic into the cache. If we see invalid errors when trying to write rows to the database, we call `updateSchema` using the most recent message schema that we've received for that table and topic.

In BigQuery, any fields that are added to a schema [must be nullable or repeated](https://cloud.google.com/bigquery/docs/tables#update-schema). That means we should only try to update the schema when trying to write a message that has new fields.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -150,6 +153,10 @@ public void put(Collection<SinkRecord> records) {
     for (SinkRecord record : records) {
       if (record.value() != null) {
         PartitionedTableId table = getRecordTable(record);
+        if (schemaRetriever != null) {
```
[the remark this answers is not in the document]
makearl: The schemaRetriever is optional configuration and only required when you want the connector to create the table or update the schema

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/retrieve/MemorySchemaRetriever.java
```diff
@@ -0,0 +1,56 @@
...
+
+  @Override
+  public Schema retrieveSchema(TableId table, String topic) {
+    String tableName = table.getTable();
+    Schema schema = schemaCache.get(getCacheKey(tableName, topic));
+    if (schema != null) {
+      return schema;
+    }
+
+    return SchemaBuilder.struct().build();
```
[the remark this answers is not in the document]
makearl: Yeah, by returning an empty schema the calling code will create a table without a schema. When we receive our first message and try to add it, we'll hit the invalid schema case from the original code and update the schema with the schema from the message

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -74,22 +78,32 @@ public AdaptiveBigQueryWriter(BigQuery bigQuery,
...
+      writeResponse = bigQuery.insertAll(request);
+      // Should only perform one schema update attempt; may have to continue insert attempts due to
+      // BigQuery schema updates taking up to two minutes to take effect
+      if (writeResponse.hasErrors()
+              && onlyContainsInvalidSchemaErrors(writeResponse.getInsertErrors())) {
+        attemptSchemaUpdate(tableId, topic);
       }
+    } catch (BigQueryException e) {
+      if (isTableMissingSchema(e)) {
+        attemptSchemaUpdate(tableId, topic);
```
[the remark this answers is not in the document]
makearl: If a table has no schema, it will raise a `BigQueryException` and the `writeResponse` will never be set. If the table has an incorrect schema, `bigQuery.insertAll` will return a writeResponse containing the errors encountered

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -74,22 +78,32 @@ public AdaptiveBigQueryWriter(BigQuery bigQuery,
...
+      } else {
+        throw e;
+      }
+
+      // If the table was missing its schema, we never received a writeResponse
+      writeResponse = bigQuery.insertAll(request);
```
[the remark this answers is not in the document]
makearl: If a table has no schema, it will raise a `BigQueryException` and the `writeResponse` will never be set in the try block so we need to retry inserting after adding a schema to the table

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -106,6 +120,18 @@ public void performWriteRequest(PartitionedTableId tableId,
     logger.debug("table insertion completed successfully");
   }
 
+  private boolean hasBigQueryResponseErrors(InsertAllResponse writeResponse, InsertAllRequest request) {
+    return request != null && writeResponse != null && writeResponse.hasErrors();
```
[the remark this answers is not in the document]
makearl: You're right, I'll get rid of this function. `writeResponse` should not be null

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/retrieve/MemorySchemaRetriever.java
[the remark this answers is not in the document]
mtagle: Could you leave a comment to this effect? either on or in the function?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -74,17 +80,27 @@ public AdaptiveBigQueryWriter(BigQuery bigQuery,
...
+      request = createInsertAllRequest(tableId, rows);
+      writeResponse = bigQuery.insertAll(request);
+      // Should only perform one schema update attempt; may have to continue insert attempts due to
+      // BigQuery schema updates taking up to two minutes to take effect
+      if (writeResponse.hasErrors()
+              && onlyContainsInvalidSchemaErrors(writeResponse.getInsertErrors())) {
+        attemptSchemaUpdate(tableId, topic);
+      }
+    } catch (BigQueryException e) {
+      if (isTableMissingSchema(e)) {
```
mtagle: if I understand correctly, there are two possiblities where we need to update the table schema:

1. the table exists and has a schema, but doesn't have some new cols (and thus needs to be updated)
2. the table exists but has no/an empty schema. This is new because this is only possible with the MemorySchemaRetriever.

is my understanding accurate?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -74,17 +80,27 @@ public AdaptiveBigQueryWriter(BigQuery bigQuery,
...
+        attemptSchemaUpdate(tableId, topic);
+      } else {
+        throw e;
       }
+
+      // If the table was missing its schema, we never received a writeResponse
+      writeResponse = bigQuery.insertAll(request);
```
mtagle: still seems to me like we could be doing this twice.  If this is only needed if the table was missing it's schema, shouldn't this be in the catch block?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
makearl: Yes, that's correct

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
makearl: It is in the catch block, I can move it under the `if (isTableMissingSchema(e))` to make it more readable though

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -74,16 +80,26 @@ public AdaptiveBigQueryWriter(BigQuery bigQuery,
...
+      if (writeResponse.hasErrors()
+              && onlyContainsInvalidSchemaErrors(writeResponse.getInsertErrors())) {
+        attemptSchemaUpdate(tableId, topic);
+      }
+    } catch (BigQueryException e) {
+      if (isTableMissingSchema(e)) {
+        attemptSchemaUpdate(tableId, topic);
+
+        // If the table was missing its schema, we never received a writeResponse
+        writeResponse = bigQuery.insertAll(request);
```
mtagle: Actually, I wonder if this should be reformatted a bit.

If we have a write response with invalid schema errors, then we retry the insert with an attempt count (lines 106-121) because the schema update might be delayed.  I assume this is also a possibility if there is no schema all all.

So, I feel like this insertAll should be covered in that block, and not here.

makearl: If we have no schema, the `writeResponse` will be null so we'll run into issues during the while loop's `writeResponse.hasErrors` check. Just to clarify, are you thinking we should update the while loop to handle the case where the writeResponse is null? That seems reasonable to me but want to verify before implementing

mtagle: yeah, I think the retry-after-schema-related-failure while loop should be updated to cover the case of this type of schema-related failure.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -74,24 +80,32 @@ public AdaptiveBigQueryWriter(BigQuery bigQuery,
...
     // Schema update might be delayed, so multiple insertion attempts may be necessary
     int attemptCount = 0;
-    while (writeResponse.hasErrors()) {
+    while (writeResponse == null || writeResponse.hasErrors()) {
       logger.trace("insertion failed");
-      if (onlyContainsInvalidSchemaErrors(writeResponse.getInsertErrors())) {
+      if (writeResponse == null || onlyContainsInvalidSchemaErrors(writeResponse.getInsertErrors())) {
+        // If the table was missing its schema, we never received a writeResponse
         logger.debug("re-attempting insertion");
         writeResponse = bigQuery.insertAll(request);
```
mtagle: we should have a try/catch around this, right? because we could get an exception of we have a null/empty schema? It's fine to just catch it and do nothing, but we should make sure the error doesn't percolate up if we intend to just retry.

makearl: Good catch, added!

mtagle:
Looks good!

PR #102, opened 2017-11-30
Add null BigQueryException reason check

makearl:
The `BigQueryException` reason can be null which caused a NullPointerException for us the other day


criccomini:
LGTM!

PR #103, opened 2018-01-26
Add additional logging info for diagnosing BigQueryWriter failures.

mtagle:
We've had a lot of issues diagnosing BigQueryExceptions from the writer threads, since we only see the top-level error, and not the full stack trace (and thus no info from BigQuery itself). This adds the stack trace to the logs for the DEBUG level.

Also adds an INFO level logging message to attempts to update table schemas.


C0urante:
I think you can just use the exception itself in the log message and get the same benefits as you would with checking to see if `logger.isDebugEnabled()`; i.e.:

`logger.debug("Error Task Stacktrace:", throwable);`

Issue #53, opened 2016-10-27
Deal with partial BigQuery failures more elegantly


mtagle:
BigQuery write requests can partially fail.  Some rows for a request may be successfully written, and others will not be.  If this is the case, the whole flush will be considered a failure we'll end up with a stack trace like this: 

```
Caused by: java.util.concurrent.ExecutionException: com.wepay.kafka.connect.bigquery.exception.BigQueryConnectException: table insertion failed for the following rows:
    [row index 3000]: backendError: null
    [row index 3001]: backendError: null
    [row index 3002]: backendError: null
```

Since the whole flush is considered a failure, kafka connect will rebalance and end up re-writing all the rows to BQ.  This results in duplicated rows.

This is not a huge issue (BQ views can be written to dedup duplicated rows), but, if possible, it would be nice to take advantage of the fact that some rows were successfully written and only attempt to write the unsuccessful rows.

whynick1:
Another interesting source provided by @criccomini, about how BQ internally handle deduplicate
https://cloud.google.com/blog/big-data/2017/06/life-of-a-bigquery-streaming-insert
![screen shot 2018-03-26 at 9 53 03 am](https://user-images.githubusercontent.com/11767413/37920404-9c4fb0a4-30db-11e8-9a5d-efb4a32b9b79.png)

PR #108, opened 2018-03-26
Deal with partial bigquery failure

whynick1:
Change has been made based on ticket: https://github.com/wepay/kafka-connect-bigquery/issues/53
1. Added an updateRowsBaseOnFailure method to filter out "succeeded rows" in partial failure to eliminate duplicate write during retry. (Design explain, remove row from the back of row list iterator because we delete based on index)
2. Added unit test regarding new method.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -86,7 +86,7 @@ public class BigQuerySinkTask extends SinkTask {
   }
 
   // For testing purposes only; will never be called by the Kafka Connect framework
-  BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever) {
+  public BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever) {
     this.testBigQuery = testBigQuery;
     this.schemaRetriever = schemaRetriever;
   }
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
@@ -32,6 +32,7 @@ import com.wepay.kafka.connect.bigquery.utils.PartitionedTableId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
@@ -72,14 +73,12 @@ public class AdaptiveBigQueryWriter extends BigQueryWriter {
    * Sends the request to BigQuery, then checks the response to see if any errors have occurred. If
    * any have, and all errors can be blamed upon invalid columns in the rows sent, attempts to
    * update the schema of the table in BigQuery and then performs the same write request.
-   * @param tableId The PartitionedTableId.
-   * @param rows The rows to write.
-   * @param topic The Kafka topic that the row data came from.
+   * @see BigQueryWriter#performWriteRequest(PartitionedTableId, List, String)
    */
   @Override
-  public void performWriteRequest(PartitionedTableId tableId,
-                                  List<InsertAllRequest.RowToInsert> rows,
-                                  String topic) {
+  public Map<Long, List<BigQueryError>> performWriteRequest(PartitionedTableId tableId,
+                                                               List<InsertAllRequest.RowToInsert> rows,
+                                                               String topic) {
     InsertAllResponse writeResponse = null;
     InsertAllRequest request = null;
 
@@ -113,7 +112,7 @@ public class AdaptiveBigQueryWriter extends BigQueryWriter {
           // no-op, we want to keep retrying the insert
         }
       } else {
-        throw new BigQueryConnectException(writeResponse.getInsertErrors());
+        return writeResponse.getInsertErrors();
       }
       attemptCount++;
       if (attemptCount >= AFTER_UPDATE_RETY_LIMIT) {
@@ -122,6 +121,7 @@ public class AdaptiveBigQueryWriter extends BigQueryWriter {
       }
     }
     logger.debug("table insertion completed successfully");
+    return new HashMap<>();
   }
 
   private void attemptSchemaUpdate(PartitionedTableId tableId, String topic) {
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
@@ -18,18 +18,21 @@ package com.wepay.kafka.connect.bigquery.write.row;
  */
 
 
+import com.google.cloud.bigquery.BigQueryError;
 import com.google.cloud.bigquery.BigQueryException;
 import com.google.cloud.bigquery.InsertAllRequest;
 
 import com.wepay.kafka.connect.bigquery.exception.BigQueryConnectException;
-
 import com.wepay.kafka.connect.bigquery.utils.PartitionedTableId;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Random;
+import java.util.Set;
 
 /**
  * A class for writing lists of rows to a BigQuery table.
@@ -69,10 +72,11 @@ public abstract class BigQueryWriter {
    * @param tableId The PartitionedTableId.
    * @param rows The rows to write.
    * @param topic The Kafka topic that the row data came from.
+   * @return map from failed row id to the BigQueryError.
    */
-  protected abstract void performWriteRequest(PartitionedTableId tableId,
-                                              List<InsertAllRequest.RowToInsert> rows,
-                                              String topic)
+  protected abstract Map<Long, List<BigQueryError>> performWriteRequest(PartitionedTableId tableId,
+                                                                        List<InsertAllRequest.RowToInsert> rows,
+                                                                        String topic)
       throws BigQueryException, BigQueryConnectException;
 
   /**
@@ -102,6 +106,7 @@ public abstract class BigQueryWriter {
     logger.debug("writing {} row{} to table {}", rows.size(), rows.size() != 1 ? "s" : "", table);
 
     Exception mostRecentException = null;
+    Map<Long, List<BigQueryError>> failedRowsMap = null;
 
     int retryCount = 0;
     do {
@@ -109,8 +114,20 @@ public abstract class BigQueryWriter {
         waitRandomTime();
       }
       try {
-        performWriteRequest(table, rows, topic);
-        return;
+        failedRowsMap = performWriteRequest(table, rows, topic);
+        if (failedRowsMap.isEmpty()) {
+          // table insertion completed with no reported errors
+          return;
+        } else if (isPartialFailure(rows, failedRowsMap)) {
+            logger.info("{} rows succeeded, {} rows failed", rows.size() - failedRowsMap.size(), failedRowsMap.size());
+            // update insert rows and retry in case of partial failure
+            rows = getFailedRows(rows, failedRowsMap.keySet());
+            mostRecentException = new BigQueryConnectException(failedRowsMap);
+            retryCount++;
+        } else {
+            // throw an exception in case of complete failure
+            throw new BigQueryConnectException(failedRowsMap);
+        }
       } catch (BigQueryException err) {
         mostRecentException = err;
         if (err.getCode() == INTERNAL_SERVICE_ERROR
@@ -143,6 +160,34 @@ public abstract class BigQueryWriter {
     throw new BigQueryConnectException(
         String.format("Exceeded configured %d attempts for write request", retries),
         mostRecentException);
+  }
+
+  /**
+   * Decide whether the failure is a partial failure or complete failure
+   * @param rows The rows to write.
+   * @param failedRowsMap A map from failed row index to the BigQueryError.
+   * @return isPartialFailure.
+   */
+  private boolean isPartialFailure(List<InsertAllRequest.RowToInsert> rows, Map<Long, List<BigQueryError>> failedRowsMap) {
+    return failedRowsMap.size() < rows.size();
+  }
+
+  /**
+   * Filter out succeed rows, and return a list of failed rows.
+   * @param rows The rows to write.
+   * @param failRowsSet A set of failed row index.
+   * @return A list of failed rows.
+   */
+  private List<InsertAllRequest.RowToInsert> getFailedRows(List<InsertAllRequest.RowToInsert> rows,
+                                                           Set<Long> failRowsSet) {
+    List<InsertAllRequest.RowToInsert> failRows = new ArrayList<>();
+    for (int index = 0; index < rows.size(); index++) {
+      if (failRowsSet.contains((long)index)) {
+        logger.debug("[row with index {}]: failed", index);
+        failRows.add(rows.get(index));
+      }
+    }
+    return failRows;
   }
 
   /**
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
@@ -19,18 +19,19 @@ package com.wepay.kafka.connect.bigquery.write.row;
 
 
 import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryError;
 import com.google.cloud.bigquery.InsertAllRequest;
 import com.google.cloud.bigquery.InsertAllResponse;
 
 import com.wepay.kafka.connect.bigquery.config.BigQuerySinkTaskConfig;
-import com.wepay.kafka.connect.bigquery.exception.BigQueryConnectException;
-
 import com.wepay.kafka.connect.bigquery.utils.PartitionedTableId;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * A simple BigQueryWriter implementation. Sends the request to BigQuery, and throws an exception if
@@ -52,16 +53,14 @@ public class SimpleBigQueryWriter extends BigQueryWriter {
   }
 
   /**
-   * Sends the request to BigQuery, and throws an exception if any errors occur as a result of doing
-   * so.
-   * @param tableId The PartitionedTableId.
-   * @param rows The rows to write.
-   * @param topic The Kafka topic that the row data came from (ignored).
+   * Sends the request to BigQuery, and return a map of insertErrors in case of partial failure.
+   * Throws an exception if any other errors occur as a result of doing so.
+   * @see BigQueryWriter#performWriteRequest(PartitionedTableId, List, String)
    */
   @Override
-  public void performWriteRequest(PartitionedTableId tableId,
-                                  List<InsertAllRequest.RowToInsert> rows,
-                                  String topic) {
+  public Map<Long, List<BigQueryError>> performWriteRequest(PartitionedTableId tableId,
+                                                            List<InsertAllRequest.RowToInsert> rows,
+                                                            String topic) {
     InsertAllRequest request = createInsertAllRequest(tableId, rows);
     InsertAllResponse writeResponse = bigQuery.insertAll(request);
     if (writeResponse.hasErrors()) {
@@ -70,9 +69,10 @@ public class SimpleBigQueryWriter extends BigQueryWriter {
           + "{}=true in the properties file",
           BigQuerySinkTaskConfig.SCHEMA_UPDATE_CONFIG
       );
-      throw new BigQueryConnectException(writeResponse.getInsertErrors());
+      return writeResponse.getInsertErrors();
     } else {
       logger.debug("table insertion completed with no reported errors");
+      return new HashMap<>();
     }
   }
 }
--- a/kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
+++ b/kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
@@ -1,0 +1,216 @@
+package com.wepay.kafka.connect.bigquery.write.row;
+
+/*
+ * Copyright 2016 WePay, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryError;
+import com.google.cloud.bigquery.InsertAllRequest;
+import com.google.cloud.bigquery.InsertAllResponse;
+import com.wepay.kafka.connect.bigquery.BigQuerySinkTask;
+import com.wepay.kafka.connect.bigquery.SinkTaskPropertiesFactory;
+import com.wepay.kafka.connect.bigquery.config.BigQuerySinkConfig;
+import com.wepay.kafka.connect.bigquery.config.BigQuerySinkTaskConfig;
+import com.wepay.kafka.connect.bigquery.exception.BigQueryConnectException;
+import org.apache.kafka.connect.data.Schema;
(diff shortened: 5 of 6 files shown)


whynick1:
@mtagle please review, thanks

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -89,6 +90,24 @@ protected InsertAllRequest createInsertAllRequest(PartitionedTableId tableId,
...
+   * @param rows The rows to write.
+   * @param failedRows The set of indexes of row that failed to write.
+   */
+  protected void updateRowsBaseOnFailure(List<InsertAllRequest.RowToInsert> rows, Set<Long> failedRows) {
+    logger.debug("updating rows to include only failed rows during write operation");
+    ListIterator<InsertAllRequest.RowToInsert> iter = rows.listIterator(rows.size());
+    while (iter.hasPrevious()) {
+      if (!failedRows.contains((long)iter.previousIndex())) {
+        iter.previous();
+        iter.remove();
```
mtagle: Thinking about this more, although I don't see any problems right now with modifying `rows` underneath us, I am sort of worried it could cause unexpected problems down the line.

I like the idea of re-using the existing retry logic a lot, but I'm wondering if there is a good way to do it without in-place modifying our row list.

whynick1:
Do you mean "using another list to store failed rows" instead of "in-place modification". I agree it's more robust. But I don't see a way to do that for following reasons:
1. assume we already have another list with filtered row, we cannot modified row with "rows = filteredRows" within performWriteRequest() method, as object in Java is passed by value
https://github.com/wepay/kafka-connect-bigquery/blob/master/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java#L61-L77

2. second choice I can think of is to "use method return", like "performWriteRequest(table, getFilteredRows(rows, failedRowIndexes), topic);". 
https://github.com/wepay/kafka-connect-bigquery/blob/master/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java#L112
However, I can't get failedRowIndexes from BigQueryConnectException because it serialize Map<index, List<bq_error>> into string.
https://github.com/wepay/kafka-connect-bigquery/blob/master/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/exception/BigQueryConnectException.java#L45-L47

Let me know what you think : )

mtagle:
My first thought is that we could modify the return type of performWriteRequest to return the rows that failed. BQ itself is already throwing exceptions if the whole request fails, but not if there is a partial failure, so we could mimic that behavior.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -69,17 +69,17 @@ private boolean isTableMissingSchema(BigQueryException e) {
   }
 
   /**
-   * Sends the request to BigQuery, then checks the response to see if any errors have occurred. If
```
mtagle: We definitely don't want to delete this. This documentation blurb is specific to AdaptiveBigQueryWriter.performWriteRequest, not performWriteRequest in general.

So we don't have to repeat a bunch of stuff, maybe instead of individually documenting each @param, maybe just do a @see link to the abstract method in the BQWriter base class.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -69,10 +72,11 @@ public BigQueryWriter(int retries, long retryWaitMs) {
    * @param tableId The PartitionedTableId.
    * @param rows The rows to write.
    * @param topic The Kafka topic that the row data came from.
+   * @return the list of failed rows in partial failure.
```
mtagle: map from failed row id to the BigQueryError

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -102,15 +106,25 @@ public void writeRows(PartitionedTableId table,
...
-        performWriteRequest(table, rows, topic);
-        return;
+        partialFailRowsMap = performWriteRequest(table, rows, topic);
+        if (partialFailRowsMap == null || partialFailRowsMap.isEmpty()) {
+          // table insertion completed with no reported errors
+          return;
+        } else {
+          // update insert rows based on failed rows in case of partial failure
+          // use partialFailRowsMap to track error instead of mostRecentException
+          rows = getFailRows(rows, partialFailRowsMap.keySet());
```
mtagle: lets have a logging message in here saying that so many rows succeeded and so many rows failed for table foo, and that we are retrying. Like the logging messages when we get a BigQueryException.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -140,9 +154,31 @@ public void writeRows(PartitionedTableId table,
...
+    }
+  }
+
+  /**
+   * Filter out succeed rows, and return a list of failed rows.
+   * @param rows The rows to write.
+   * @param partialFailRowsIndex A set of failed row index
+   * @return A list of failed rows
+   */
+  private List<InsertAllRequest.RowToInsert> getFailRows(List<InsertAllRequest.RowToInsert> rows,
```
mtagle: getFailedRows

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
```diff
@@ -52,16 +52,17 @@ public SimpleBigQueryWriter(BigQuery bigQuery, int retry, long retryWait) {
...
 
   /**
-   * Sends the request to BigQuery, and throws an exception if any errors occur as a result of doing
-   * so.
+   * Sends the request to BigQuery, and return a map of insertErrors in case of partial failure.
+   * Throws an exception if any other errors occur as a result of doing so.
    * @param tableId The PartitionedTableId.
    * @param rows The rows to write.
    * @param topic The Kafka topic that the row data came from (ignored).
+   * @return the list of failed rows in case of partial failure, o/w return null.
```
mtagle: otherwise over o/w. Or maybe "or null if all rows succeeded"

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
```diff
@@ -52,16 +52,17 @@ public SimpleBigQueryWriter(BigQuery bigQuery, int retry, long retryWait) {
   }
 
   /**
-   * Sends the request to BigQuery, and throws an exception if any errors occur as a result of doing
-   * so.
+   * Sends the request to BigQuery, and return a map of insertErrors in case of partial failure.
```
mtagle: technically we may return insertErrors for kinds of errors other than partial failures. Trying to data into a table where the schema doesn't match, for example.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -140,9 +154,31 @@ public void writeRows(PartitionedTableId table,
         }
       }
     } while (retryCount <= retries);
-    throw new BigQueryConnectException(
-        String.format("Exceeded configured %d attempts for write request", retries),
-        mostRecentException);
+
+    if (mostRecentException == null) {
+      // indicate the most recent exception is partial write failure
+      throw new BigQueryConnectException(partialFailRowsMap);
```
mtagle: We still want to say that we exceeded our configured retry attempts.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -102,15 +106,25 @@ public void writeRows(PartitionedTableId table,
...
         waitRandomTime();
       }
       try {
-        performWriteRequest(table, rows, topic);
-        return;
+        partialFailRowsMap = performWriteRequest(table, rows, topic);
+        if (partialFailRowsMap == null || partialFailRowsMap.isEmpty()) {
+          // table insertion completed with no reported errors
+          return;
+        } else {
```
mtagle: We may also want to go through the failed rows and make sure that the error is actually a partial failure error. We can just straight up throw an exception if it's something else, rather than retrying probably futilely.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
whynick1: make a lot of sense

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: good catch

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: It's not easy to tell a "partial failure" based on `partialFailRowsMap, because there isn't an "error code" in `BigQueryError` object. 
Actually, I think if `bigQuery.insertAll()` does not throw a BigQueryException, meanwhile `writeResponse.hasErrors()`, it is a "partial fail" (consistent with previous stack trace).
What you think?
https://github.com/wepay/kafka-connect-bigquery/pull/108/files#diff-5e819bb0e646d1ad0aa8d0281013ef43R67
https://github.com/wepay/kafka-connect-bigquery/issues/53

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: sure we can do that

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: sure

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: sure

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
whynick1: sure

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
mtagle: No. Anything that has a non-200 error code will be an exception. (This is because the bigquery client is just a thin wrapper over a rest interface.)

Take a look at the "Success HTTP response codes" section in that troubleshooting page to see the types of "errors" that will not throw but instead return an insertResponse with errors.

This line seems particularly useful:
> Rows that did not have a schema mismatch will have an error with the reason property set to stopped, and can be re-sent as-is.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
mtagle: I think my comment here: https://github.com/wepay/kafka-connect-bigquery/pull/108#discussion_r178650979 addresses this.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
whynick1: Now I am clear! Thanks!

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -145,6 +163,35 @@ public void writeRows(PartitionedTableId table,
         mostRecentException);
   }
 
+  /**
+   * Decide whether the failure is a partial failure or complete failure
+   * @param numOfFailedRows Number of rows that failed
+   * @param numOfAllRows Number of rows to insert
+   * @return isPartialFailure
+   */
+  private boolean isPartialFailure(int numOfFailedRows, int numOfAllRows) {
```
mtagle: I'd prefer this take in the list of attempted rows and list of failed rows, in case we want to make the check more nuanced later.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -145,6 +163,35 @@ public void writeRows(PartitionedTableId table,
...
+
+  /**
+   * Filter out succeed rows, and return a list of failed rows.
+   * @param rows The rows to write.
+   * @param failRowsIndex A set of failed row index
+   * @return A list of failed rows
+   */
+  private List<InsertAllRequest.RowToInsert> getFailedRows(List<InsertAllRequest.RowToInsert> rows,
+                                                           Set<Long> failRowsIndex) {
+    logger.info("{} rows succeeded, {} rows failed", rows.size() - failRowsIndex.size(), failRowsIndex.size());
```
mtagle: I think this logging message should be in writeRows itself.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -145,6 +163,35 @@ public void writeRows(PartitionedTableId table,
...
+    List<InsertAllRequest.RowToInsert> failRows = new ArrayList<>();
+    for (int index = 0; index < rows.size(); index++) {
+      if (failRowsIndex.contains((long)index)) {
+        logger.info("Adding [row index %d] for retry", index);
```
mtagle: This sounds like it's going to be REALLY verbose. This seems like potentially useful data, but maybe it should be at the debug level instead of info?

Also, this method is just filtering the rows. We don't know what the caller will use that data for. Instead of saying that we're going to retry, maybe just say the row with that index failed?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -69,10 +72,11 @@ public BigQueryWriter(int retries, long retryWaitMs) {
    * @param tableId The PartitionedTableId.
    * @param rows The rows to write.
    * @param topic The Kafka topic that the row data came from.
+   * @return map from failed row id to the BigQueryError or null if all rows succeeded.
```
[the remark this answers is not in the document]
whynick1: sure

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: got it

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: Sure, my previous concern if we log somewhere else, we have to iterate through the list again, which might be expensive. While partial failure is not very common, I think you are right.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
whynick1: absolutely

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
```diff
@@ -0,0 +1,79 @@
...
+        logger = LoggerFactory.getLogger(BigQueryWriterTest.class);
+        testTableId = new PartitionedTableId.Builder("testDataset", "testBaseTable").build();
+        testRows = new ArrayList<>();
+        testRows.add(mock(InsertAllRequest.RowToInsert.class));
+        testRows.add(mock(InsertAllRequest.RowToInsert.class));
+        testTopic = "testTopic";
+    }
+
+    @Test(expected = BigQueryConnectException.class)
+    public void testWriteRowsShouldThrowExcepiton() throws InterruptedException {
```
mtagle: write rows should throw exception if...? Can you elaborate in a comment a bit more detail on what this is testing? (also excepiton -> exception)

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+        BigQueryWriter mockBigQueryWriter = mock(SimpleBigQueryWriter.class);
+        //mimic partial failure
+        when(mockBigQueryWriter.performWriteRequest(testTableId, testRows, testTopic)).thenReturn(testPartialFailRowsMap);
+        doCallRealMethod().when(mockBigQueryWriter).writeRows(testTableId, testRows, testTopic);
+        mockBigQueryWriter.writeRows(testTableId, testRows, testTopic);
+    }
+
+    @Test
+    public void testWriteRowsShouldPass() throws InterruptedException {
```
mtagle: similarly here.

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
whynick1: I want to test this 'if ... else if ... else' block here: https://github.com/wepay/kafka-connect-bigquery/pull/108/commits/3d51fb6e6edf7b7811c3ca785437c410ab998bec#diff-f00fa7ac2c3aa6ac09bf9381986fc494R118
```
if succeed, return;
else failed
    if partial fail, update rows, (will throw BigQueryConnectionException once retry == 0)
    else throw BigQueryConnectionException
```
Currently, I only covers "succeed" and 'failed' cases, though it would be ideal to cover 3 cases. I am not sure how to discern last 2 case in unit test, any suggestion?

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.mockito.Mockito.*;
+
+public class BigQueryWriterTest {
+    private Logger logger;
```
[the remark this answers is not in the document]
whynick1: got it

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
```diff
@@ -0,0 +1,77 @@
...
+    private PartitionedTableId testTableId ;
```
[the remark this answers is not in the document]
whynick1: got it

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
whynick1: 👍 great idea!

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
```diff
@@ -0,0 +1,176 @@
...
+    public void testBigQueryPartialFailure() {
+        final String topic = "test_topic";
+        final String dataset = "scratch";
+        final Map<String, String> properties = makeProperties("3", "2000", topic, dataset);
+
+        Map<Long, List<BigQueryError>> insertErrorMap = mock(Map.class);
+        when(insertErrorMap.isEmpty()).thenReturn(false);
+        when(insertErrorMap.size()).thenReturn(1);
+
+        InsertAllResponse insertAllResponseWithError = mock(InsertAllResponse.class);
```
mtagle: maybe some comments here would be good, like "first attempt (partial failure)" "second attempt (success)"

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        SinkTaskContext sinkTaskContext = mock(SinkTaskContext.class);
+
+        BigQuerySinkTask testTask = new BigQuerySinkTask(bigQuery, null);
+        testTask.initialize(sinkTaskContext);
+        testTask.start(properties);
+        testTask.put(sinkRecordList);
+        testTask.flush(Collections.emptyMap());
+
+        verify(bigQuery, times(2)).insertAll(anyObject());
```
mtagle: I'd also want to verify that the second insertAll was called with just the failed rows.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -86,7 +86,7 @@ public BigQuerySinkTask() {
   }
 
   // For testing purposes only; will never be called by the Kafka Connect framework
-  BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever) {
+  public BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever) {
```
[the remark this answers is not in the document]
whynick1: yes, because of different package and non-subclass

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -122,6 +120,7 @@ public void performWriteRequest(PartitionedTableId tableId,
       }
     }
     logger.debug("table insertion completed successfully");
+    return writeResponse.getInsertErrors();
```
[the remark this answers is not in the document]
whynick1: sure, I will use a empty list

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/SimpleBigQueryWriter.java
```diff
@@ -70,9 +68,10 @@ public void performWriteRequest(PartitionedTableId tableId,
           + "{}=true in the properties file",
           BigQuerySinkTaskConfig.SCHEMA_UPDATE_CONFIG
       );
-      throw new BigQueryConnectException(writeResponse.getInsertErrors());
+      return writeResponse.getInsertErrors();
     } else {
       logger.debug("table insertion completed with no reported errors");
+      return writeResponse.getInsertErrors();
```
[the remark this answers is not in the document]
whynick1: agree

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
whynick1: sure

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
```diff
@@ -0,0 +1,176 @@
...
+     * @param bigqueryRetry The number of retries.
+     * @param bigqueryRetryWait The wait time for each retry.
+     * @param topic The topic of the record.
+     * @param dataset The dataset of the record.
+     * @return The map of bigquery sink configurations.
+     */
+    private Map<String,String> makeProperties(String bigqueryRetry, String bigqueryRetryWait, String topic, String dataset) {
+        Map<String, String> properties = propertiesFactory.getProperties();
+        properties.put(BigQuerySinkTaskConfig.BIGQUERY_RETRY_CONFIG, "3");
+        properties.put(BigQuerySinkTaskConfig.BIGQUERY_RETRY_WAIT_CONFIG, "2000");
```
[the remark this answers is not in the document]
whynick1: ok

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriterTest.java
whynick1: I can do that

PR #111, opened 2018-04-11
Handle date type in integration test

whynick1:
Added BQ 'DATE' type handle to `convertField()` function, so that the test `testLogicalTypes()` would pass

### DIFF
--- a/kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/BigQueryConnectorIntegrationTest.java
+++ b/kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/BigQueryConnectorIntegrationTest.java
@@ -37,6 +37,9 @@ import org.junit.Test;
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 
+import java.time.LocalDate;
+import java.time.ZoneOffset;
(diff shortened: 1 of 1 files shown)


kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/BigQueryConnectorIntegrationTest.java
```diff
@@ -263,9 +273,9 @@ public void testLogicalTypes() {
     // {"row": 1, "timestamp-test": 0, "date-test": 0}
     expectedRows.add(Arrays.asList(1L, 0L, 0L));
     // {"row": 2, "timestamp-test": 42000000, "date-test": 4200}
-    expectedRows.add(Arrays.asList(2L, 42000000000L, 362880000000000L));
+    expectedRows.add(Arrays.asList(2L, 42000000000L, 362880000000L));
```
mtagle: do we understand why this changed?

mtagle: Actually I think I have some memory of changing/fixing the way that timestamps were dealt with at some point, so I'm going to say that this is good.

PR #119, opened 2018-06-26
Add executor shutdown when BigQuerySinkTask is stopped

jgao54:


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -79,6 +69,7 @@
   private TopicPartitionManager topicPartitionManager;
 
   private KCBQThreadPoolExecutor executor;
+  private static final int executor_shutdown_timeout_sec = 30;
```
jgao54: @mtagle , not sure if we wanted to make this a config or not. It's kind of arbitrary right now, reason I didn't add is yet is because it's used for log.warn only, but existing tasks continue to run even after shutdown is signaled so there's no material impact afaik.

criccomini: Nit: Should be EXECUTOR_SHUTDOWN_TIMEOUT_SEC

criccomini: IMO, 30s is fine as-is here.

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTaskTest.java
```diff
@@ -384,11 +381,36 @@ public void testVersion() {
...
+
+    BigQuerySinkTask testTask = new BigQuerySinkTask(bigQuery, null);
+    testTask.initialize(sinkTaskContext);
+    testTask.start(properties);
+    testTask.put(Collections.singletonList(spoofSinkRecord(topic)));
+
+    KCBQThreadPoolExecutor executor = testTask.getExecutor();
+    assertFalse((executor.isTerminating() || executor.isTerminated()));
+    testTask.stop();
+    assertTrue((executor.isTerminating() || executor.isTerminated()));
```
criccomini: Curious about this. Shouldn't isTerminating always be false and isTerminated always be true here?

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTaskTest.java
```diff
@@ -384,11 +381,36 @@ public void testVersion() {
     assertNotNull(new BigQuerySinkTask().version());
   }
 
-  // Doesn't do anything at the moment, but having this here will encourage tests to be written if
```
criccomini: lol :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -241,7 +232,20 @@ public void start(Map<String, String> properties) {
...
+    } catch (InterruptedException e) {
+      logger.warn("{} active threads are still executing tasks {}s after shutdown is signaled.",
+          executor.getActiveCount(), executor_shutdown_timeout_sec);
+    } finally {
+      logger.trace("task.stop()");
+    }
+  }
+
+  @VisibleForTesting
+  KCBQThreadPoolExecutor getExecutor() {
```
criccomini: This is bugging me. Using dependency injection and taking a factory or lambda in the constructor also bugs me, but at least doesn't introduce a this Google dependency.

Another idea would be to just expose getActiveCount as a public method, and use that in the tests as a proxy for isTerminated.

mtagle: Yeah, I'm not a fan of this either. Making getActiveCount public makes sense to me.

jgao54: Okay, alternatively I could also use java reflection if we want to keep the class clean.

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTaskTest.java
jgao54: ah yeah, you are right. Forgot to change this. I wrote the test first and then added the `awaitTermination` part in `stop` :)

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTaskTest.java
C0urante: #planningahead

jgao54: @C0urante  thanks!

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -85,6 +85,7 @@ public BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever)
   @Override
   public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
     try {
+      logger.info("Flushing {} records in the sink.", executor.getActiveCount());
```
mtagle: is this supposed to be executor.getActiveCount()?

jgao54: @mtagle Ah, I agree it's a bit vague and potentially confusing. (active + completed) tasks is incorrect since it would return a count of all tasks completed by this executor; and it's not possible to just get the "interval" of tasks since the last flush. I'm going to remove this line.

PR #120, opened 2018-06-27
Added batch config

stoynov96:
Added a batch upload config option to specify a list of topics to be batch loaded through GCS. Validation that those topics are in the overall list of topics is not performed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,6 @@ build/
 key.json
 
 test.conf
-kcbq-connector/src/integration-test/resources/test.properties
+test.properties
 kcbq-connector/test/docker/connect/properties/
 kcbq-connector/out/
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,6 @@ build/
 key.json
 
 test.conf
-test.properties
+kcbq-connector/src/integration-test/resources/test.properties
 kcbq-connector/test/docker/connect/properties/
 kcbq-connector/out/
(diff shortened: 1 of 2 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -59,6 +60,13 @@
   private static final String TOPICS_DOC =
       "A list of Kafka topics to read from";
 
+  public static final String ENABLE_BATCH_CONFIG =                         "enableBatchLoad";
+  public static final ConfigDef.Type ENABLE_BATCH_TYPE =                   ConfigDef.Type.LIST;
+  public static final List<String> ENABLE_BATCH_DEFAULT =                  Collections.emptyList();
+  public static final ConfigDef.Importance ENABLE_BATCH_IMPORTANCE =       ConfigDef.Importance.LOW;
+  public static final String ENABLE_BATCH_DOC =
+      "A list of Kafka topics to batch load through GCS";
```
mtagle: We should mention we expect this to be a sub-list of the full topics list.
Maybe something like
"The sublist of topics to be batch loaded through GCS"?

.gitignore
```diff
@@ -22,6 +22,6 @@ build/
 key.json
 
 test.conf
-kcbq-connector/src/integration-test/resources/test.properties
+test.properties
```
mtagle: Did this need to be changed here? Just in the interest of keeping these PRs limited to the GCS stuff, we probably don't want this change here, unless I'm missing something.

PR #130, opened 2018-07-18
Finished new table writer, implemented batch load flow

stoynov96:
Finished the new table writer and altered it to work similarly to the normal table writer.
Introduced a GCS writer.
Added a configuration for the name of the bucket where GCS blobs should be uploaded.
Altered the flow of BigQuerySinkTask to batch process records in topics configured to be batch loaded

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
@@ -47,8 +47,8 @@ import java.util.Random;
 /**
  * A class for batch writing list of rows to BigQuery through GCS
  */
-public class GCSWriter {
-  private static final Logger logger = LoggerFactory.getLogger(GCSWriter.class);
+public class GCSToBQWriter {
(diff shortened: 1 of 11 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
```diff
@@ -0,0 +1,178 @@
...
+          new ByteArrayInputStream(toJson(rows).getBytes("UTF-8")), blobInfo);
+      return resultBlob;
+    } catch (UnsupportedEncodingException uee) {
+      throw new GCSConnectException("Failed to upload blob to GCS", uee);
+    }
+  }
+
+  private Blob uploadBlobToGcs(InputStream blobContent, BlobInfo blobInfo) {
+    // todo look into creating from a string because this is depreciated - input stream cannot retry
+    // todo consider if it would be worth it to switch to a resumable method of uploading
```
mtagle: Can you elaborate on this? I'm fine with committing this as is, but we should probably open an issue with more information this if so so it's easier to track down and change later if we decide it's needed.

PR #137, opened 2018-09-04
Separate gcs to bq load

mtagle:
Add a separate runnable to run the GCS to BQ load job trigger.

BQ load jobs have quotas, so running a load job every chance we get results in hitting those quotas almost immediately. Also, since triggering a job just triggers a job, without keeping track of the state of those jobs we end up in a situation where those loads were failing silently (due to the aforementioned quotas). 

To keep ourselves under the quotas, we're having one task run an executor service that runs GCS -> BQ load jobs for each configured table every 2 minutes. This runnable grabs all the blobs it can (that aren’t already being loading by an existing job) for each table and runs a single load job containing all of those blobs. We also keep track of jobs and see if they failed or succeeded. We delete the relevant blobs once a task completes.

Extra note: so that the load job trigger knows what blobs belong in what tables, the initial load into GCS now add metadata to the blobs that says what table they should be uploaded to.


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+
+  private final BigQuery bigQuery;
+  private final Bucket bucket;
+  private final Map<Job, List<Blob>> activeJobs;
+  private final Set<Blob> claimedBlobs;
+
+  // these numbers are intended to try to make this task not excede Google Cloud Quotas.
+  // see: https://cloud.google.com/bigquery/quotas#load_jobs
+  private static int FILE_LOAD_LIMIT = 10000;
+  // max number of files we can load in a single load job
```
criccomini: i think this comment and the one below are referring to the variables above them, which is a bit confusing.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+  public GCSToBQLoadRunnable(BigQuery bigQuery, Bucket bucket) {
+    this.bigQuery = bigQuery;
+    this.bucket = bucket;
+    this.activeJobs = new HashMap<>();
+    this.claimedBlobs = new HashSet<>();
+  }
+
+  private Map<TableId, List<Blob>> getBlobsUpToLimit() {
+    Map<TableId, List<Blob>> tableToURIs = new HashMap<>();
+    //List<String> URIs = new ArrayList<>();
```
criccomini: Delete?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+      logger.error("Found blob {}\\{} with no metadata.", blob.getBucket(), blob.getName());
+      return null;
+    }
+
+    String serializedTableId = blob.getMetadata().get(GCSToBQWriter.GCS_METADATA_TABLE_KEY);
+    Matcher matcher = METADATA_TABLE_PATTERN.matcher(serializedTableId);
+    String project = matcher.group("project");
+    String dataset = matcher.group("dataset");
+    String table = matcher.group("table");
+
```
criccomini: can you add a debug statement here so we can see what project/dataset/id are for a given blob if we need to

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+   * @param bucket the the GCS bucket to read from.
+   */
+  public GCSToBQLoadRunnable(BigQuery bigQuery, Bucket bucket) {
+    this.bigQuery = bigQuery;
+    this.bucket = bucket;
+    this.activeJobs = new HashMap<>();
+    this.claimedBlobs = new HashSet<>();
+  }
+
+  private Map<TableId, List<Blob>> getBlobsUpToLimit() {
```
criccomini: Javadoc

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+
+      if (!tableToURIs.containsKey(table)) {
+        // initialize maps, if we haven't seen this table before.
+        tableToURIs.put(table, new ArrayList<>());
+        tableToCurrentLoadSize.put(table, 0L);
+      }
+
+      long newSize = tableToCurrentLoadSize.get(table) + blob.getSize();
+      // if this file does not cause us to exceed our per-request quota limits...
+      if (newSize <= MAX_LOAD_SIZE_B && tableToURIs.get(table).size() < FILE_LOAD_LIMIT) {
```
criccomini: Paranoia: not sure if <= or < is appropriate, so I vote for < when cehcking MAX_LOAD_SIZE_B

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+    Job job = bigQuery.create(JobInfo.of(loadJobConfiguration));
+    // update active jobs and claimed blobs.
+    activeJobs.put(job, blobs);
+    claimedBlobs.addAll(blobs);
+    logger.info("Triggered load job for table {} with {} blobs.", table, blobs.size());
+    return job;
+  }
+
+  private List<String> blobsToURIs(List<Blob> blobs) {
+    List<String> uris = new ArrayList<>(blobs.size());
```
criccomini: The functional programmer in me really just wants to use blobs.stream().map().collect() for this. Since it's Java, not sure if it's going to look better or not, but worth checking.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+    return newJobs;
+  }
+
+  private Job triggerBigQueryLoadJob(TableId table, List<Blob> blobs) {
+    List<String> uris = blobsToURIs(blobs);
+    // create job load configuration
+    LoadJobConfiguration loadJobConfiguration =
+        LoadJobConfiguration.newBuilder(table, uris)
+            .setFormatOptions(FormatOptions.json())
+            .setCreateDisposition(JobInfo.CreateDisposition.CREATE_IF_NEEDED)
```
criccomini: According to https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs the default write disposition is WRITE_IF_EMPTY. I think we might need to set write disposition to WRITE_APPEND. Is it possible to do this in the builder?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -0,0 +1,232 @@
...
+    while (jobIterator.hasNext()) {
+      Job job = jobIterator.next();
+      try {
+        if (job.isDone()) {
+          List<Blob> blobsToDelete = activeJobs.remove(job);
+          successCount++;
+          for (Blob blob : blobsToDelete) {
+            try {
+              blob.delete();
+              claimedBlobs.remove(blob);
```
criccomini: Wondering a little bit about failure scenarios. If a thread fails for some reason (e.g. delete() fails), activeJobs would be mutated but claimedBlobs would not. Is that an problem?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
C0urante: DOOO ITTTTT

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -83,33 +87,32 @@
...
    * For testing purposes only; will never be called by the Kafka Connect framework.
    *
    * @param testBigQuery {@link BigQuery} to use for testing (likely a mock)
    * @param schemaRetriever {@link SchemaRetriever} to use for testing (likely a mock)
-   * @param testGCS {@link Storage} to use for testing (likely a mock)
+   * @param testGcs {@link Storage} to use for testing (likely a mock)
    * @see BigQuerySinkTask#BigQuerySinkTask()
    */
-  public BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever, Storage testGCS) {
+  public BigQuerySinkTask(BigQuery testBigQuery, SchemaRetriever schemaRetriever, Storage testGcs) {
```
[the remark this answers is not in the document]
mtagle: I might be able to, but this is already the way it was. I just changed the name of the variable to be more consistent.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -274,6 +277,11 @@ private GCSToBQWriter getGCSWriter() {
   @Override
   public void start(Map<String, String> properties) {
     logger.trace("task.start()");
+    boolean hasGCSBQTask = false;
```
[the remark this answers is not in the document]
mtagle: good call, but I realize I can actually do a bit better.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
mtagle: noted, I'll fix that.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
mtagle: fair enough.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
mtagle: I also wanted to, but I'm just really non-fluent in functional java. So, sure :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
mtagle: yep. (and that seems like an oversight, so ty for pointing it out)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
mtagle: After thinking about this for a bit: When I wrote this I was imagining blobs that we couldn't delete as permanent errors (a permission issue, for example), but I realize that's not true. We could definitely get an error from a delete that was transient. So I'll modify this.

PR #138, opened 2018-09-06
Fix tar creation

mtagle:
Fix tar creation in two ways:
1. previously tar creation took individual tars and squished them together. This resulted in some weirdness when individual projects would pull in two different versions of the same dependency, as the resulting overall tar would end up having those two different versions in it (rather than just one).
2. Pulling in debezium was resulting in a tar that used the latest kafka (2.0.0 at the moment). Add some gradle configuration to force kafka resolution to 1.1.1.

Fix readme and integration tests to use new tar build command/location.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ $ ./gradlew clean distTar
 And then extract its contents:
 
 ```bash
-$ mkdir bin/jar/ && tar -C bin/jar/ -xf kcbq-confluent/build/distributions/kafka-connect-bigquery-*.tar
+$ mkdir bin/jar/ && tar -C bin/jar/ -xf kcbq-confluent/build/distributions/kcbq-confluent-*.tar
 ```
 
 ### Setting-Up Background Processes
--- a/README.md
+++ b/README.md
@@ -49,13 +49,13 @@ $ cd /path/to/kafka-connect-bigquery/
 Begin by creating a tarball of the connector with the Confluent Schema Retriever included:
 
 ```bash
-$ ./gradlew clean confluentTarBall
+$ ./gradlew clean distTar
 ```
 
 And then extract its contents:
 
 ```bash
-$ mkdir bin/jar/ && tar -C bin/jar/ -xf bin/tar/kcbq-connector-*-confluent-dist.tar
+$ mkdir bin/jar/ && tar -C bin/jar/ -xf kcbq-confluent/build/distributions/kafka-connect-bigquery-*.tar
 ```
 
 ### Setting-Up Background Processes
--- a/build.gradle
+++ b/build.gradle
@@ -7,24 +7,6 @@ allprojects {
 // for some reason this does not work if applied in allprojects.
 apply plugin: 'distribution'
 
-configurations.all {
-
-    /*
-    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
-        if (details.requested.group == 'org.apache.kafka') {
-            details.useVersion '1.1.1'
-            // force our kafka version.
-        }
-    }
-    */
-    resolutionStrategy {
-        // none of this seems to be working =/
-        force 'org.apache.kafka:connect-api:1.1.1', 'org.apache.kafka:kafka-clients:1.1.1'
-        //forcedModules = ['org.apache.kafka:connect-api:1.1.1', 'org.apache.kafka:kafka-clients:1.1.1']
-        //preferProjectModules()
-    }
-}
-
 // END ALL PROJECTS
 
 // BEGIN SUBPROJECTS //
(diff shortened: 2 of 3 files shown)


build.gradle
```diff
@@ -1,7 +1,12 @@
 // BEGIN ALL PROJECTS //
 allprojects {
     apply plugin: 'java'
+    apply plugin: 'project-report' // todo delete me
```
criccomini: delete? :)

build.gradle
```diff
@@ -1,7 +1,12 @@
...
 }
+
+// for some reason this does not work if applied in allprojects.
+apply plugin: 'distribution'
```
criccomini: Can you move this to the `// ROOT PROJECT` section?

criccomini: Actually, do we even need this? The distribution plugin is applied specifically to the kcbq-confluent package down below.

build.gradle
```diff
@@ -257,6 +240,23 @@ project('kcbq-api') {
 }
 
 project('kcbq-confluent') {
+    apply plugin: 'distribution'
+
+    configurations.all {
+        resolutionStrategy {
```
criccomini: Can you add a `// comment` explaining why

build.gradle
```diff
@@ -257,6 +240,23 @@ project('kcbq-api') {
...
+            force 'org.apache.kafka:connect-api:1.1.1', 'org.apache.kafka:kafka-clients:1.1.1'
+        }
+    }
+
+    distributions {
+        main {
+            baseName = 'kafka-connect-bigquery'
```
criccomini: nit: name kcbq-confluent .. theoretically, if we had another schema reg implementation (e.g. proto), we could then have a kcbq-proto dist package in another project as well. I prefer naming the same as the submodule.

kcbq-connector/test/integrationtest.sh
```diff
@@ -205,8 +205,9 @@ warn 'Deleting existing BigQuery test tables'
 # Executing connector in standalone mode (this is the execution portion of the actual test)
 statusupdate 'Executing Kafka Connect in Docker'
 
-# Run clean task to ensure there's only one connector tarball in the build directory
-"$GRADLEW" -q -p "$BASE_DIR/../.." clean confluentTarBall
+# Run clean task to ensure there's only one connector tarball in the build/dist directory
+# todo fixme
```
criccomini: Todo?

mtagle: oops, it's already fixed :)

kcbq-connector/test/integrationtest.sh
```diff
@@ -232,8 +232,8 @@ echo >> "$CONNECTOR_PROPS"
 CONNECT_DOCKER_IMAGE='kcbq/connect'
 CONNECT_DOCKER_NAME='kcbq_test_connect'
 
-cp "$BASE_DIR"/../../bin/tar/kcbq-connector-*-confluent-dist.tar "$DOCKER_DIR/connect/kcbq.tar"
-cp "$BASE_DIR"/../../bin/tar/kcbq-connector-*-confluent-dist.tar "$DOCKER_DIR/connect/retriever.tar"
+cp "$BASE_DIR"/../../kcbq-confluent/build/distributions/kafka-connect-bigquery-*.tar "$DOCKER_DIR/connect/kcbq.tar"
```
criccomini: I think this has to be kcbq-confluent* now

kcbq-connector/test/integrationtest.sh
```diff
@@ -232,8 +232,8 @@ echo >> "$CONNECTOR_PROPS"
...
+cp "$BASE_DIR"/../../kcbq-confluent/build/distributions/kafka-connect-bigquery-*.tar "$DOCKER_DIR/connect/retriever.tar"
```
criccomini: I think this has to be kcbq-confluent* now

README.md
```diff
@@ -49,13 +49,13 @@ $ cd /path/to/kafka-connect-bigquery/
...
 ```bash
-$ ./gradlew clean confluentTarBall
+$ ./gradlew clean distTar
 ```
 
 And then extract its contents:
 
 ```bash
-$ mkdir bin/jar/ && tar -C bin/jar/ -xf bin/tar/kcbq-connector-*-confluent-dist.tar
+$ mkdir bin/jar/ && tar -C bin/jar/ -xf kcbq-confluent/build/distributions/kafka-connect-bigquery-*.tar
```
criccomini: I think this has to be kcbq-confluent* now

build.gradle
```diff
@@ -257,6 +236,24 @@ project('kcbq-api') {
 }
 
 project('kcbq-confluent') {
+    apply plugin: 'distribution'
+
+    configurations.all {
+        resolutionStrategy {
+            force 'org.apache.kafka:connect-api:1.1.1', 'org.apache.kafka:kafka-clients:1.1.1'
+            // depending on debezium results in us pulling in the most recent kafka version, but we want to use 1.1.1.
```
criccomini: Nit: can you move this to the line above `force`?

README.md
mtagle: ah, yes, nice catch :)

PR #139, opened 2018-09-14
Fix table-metadata related issues in GCS loading.

mtagle:
A tableId with a null project will no longer be written as a stringified "null" in the metadata. (Now that metadata will just not have any project information).
Fix GCSToBQLoadRunnable's handling of parsing that information.

Also add tests to GCSToBQLoadRunnable's getTableFromBlob method.


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
```diff
@@ -116,11 +108,13 @@ public void writeRows(List<RowToInsert> rows,
     }
 
     // Create a job configuration
+    /*
```
criccomini: Any reason to keep this floating around?

criccomini:
LGTM minus the one comment nit

PR #142, opened 2018-09-28
Fix concurrent mod in gcs loader

criccomini:
PR addresses:

* Fix concurrent modification issue in delete method
* Fix concurrent modification in check method
* Speed up GCS uploads by switching from Blob to BlobId when hashing data
* Fix erroneous UpperCamelCase config
* Turn on HTML reports for findbugs


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSToBQLoadRunnable.java
```diff
@@ -124,7 +130,10 @@ public GCSToBQLoadRunnable(BigQuery bigQuery, Bucket bucket) {
         tableToURIs.get(table).add(blob);
         tableToCurrentLoadSize.put(table, newSize);
       }
+      // TODO else should we break if we've exceeded limit?
```
mtagle: todo? but my thought is we may still encounter a file that fits... in theory, at least.

criccomini: Gotcha. That makes sense. On the flip side, my concern is that list actually paginates through all of the files in the bucket, which is a slow operation--keeps calling next page, which is an HTTP request. We hit a situation where there were 100,000+ files, and the list operation was quite slow, and was a waste since we had filled the file list almost immediately. That said, it was a bug, and it's a bit of an edge case, so I'm OK removing TODO.

criccomini:
Build failed erroneously. Retriggering.

PR #144, opened 2018-10-23
Integration tests now pass!

mtagle:
Couple of classes of changes here:
- integration tests have been upgraded to the confluent dependency versions implied by the normal build versions.
- rectify some minorly incorrect information regarding how to run the integration tests in the README
- fix jar creation! kcbq-confluent was the source of the jar and was including only its dependencies in the final jar, not itself. That's been fixed.


criccomini:
LGTM--feel free to merge once Travis finishes

Issue #149, opened 2019-02-11
kafka-connect-bigquery can't connect to a schema registry protected by basic auth


Riduidel:
I'm deploying Kafka-connect big query connector with a kfka cluster deployed at aiven.io.
In this context, the schema registry is accessed on https using basic auth.
I've set the various schema registry properties in kafka config with no success.
After having taken a look at the code, it appears `SchemaRegistrySchemaRetriever#configure(...)` method creates the client using the `CachedSchemaRegistryClient(String, int)` constructor, which doesn't allow setting the basic auth headers. As a consequence, basic auth is not possible.

Riduidel:
I'll try hacking a solution to initialize the RestService using all schema registry configuration, to see how it goes. If it works, it'll be subject to a pull request. Will you accept it ?

Riduidel:
OK, since `RestService` don't either provide any wway to inject basic auth correctly, the only remaining path is to set the basic auth in url which is, to say the least, not that good.

Anyway, if you face, that issue, it is the only solution, so choose both user and password carefully, and in a way that is easy to change

Riduidel:
Aaaaaaand it's also a fail, since kafka-connect expurge the basic auth info from the schema registry https url. A colleague suggests me to use socat as an https proxy, and that's the direction I'll take

PR #150, opened 2019-02-11
[ISSUE 149] Make underlying Schema Registry client configurable

C0urante:
Attempts to address https://github.com/wepay/kafka-connect-bigquery/issues/149

NOTE: This has not been tested

@Riduidel could you take a look and see if this is an option? Using the source branch for this PR, you can now configure the underlying client used by the `SchemaRegistrySchemaRetriever` directly, using something like:
```properties
schemaRegistryClient.basic.auth.credentials.source=SASL_INHERIT
schemaRegistryClient.basic.auth.info=username:password
```

The configs currently supported by the client can be found [here](https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/client/SchemaRegistryClientConfig.java).

@criccomini @mtagle saw the issue this morning, decided to sketch this out. Don't have time to test it but it's probably a start in the right direction if you guys are interested in adding this kind of security support to the `SchemaRegistrySchemaRetriever`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,7 +12,7 @@ project.ext {
     googleCloudVersion = '0.47.0-alpha'
     googleAuthVersion = '0.9.0'
     googleCloudGsonVersion = '2.8.5'
-    ioConfluentVersion = '3.2.0'
+    ioConfluentVersion = '5.1.1'
     junitVersion = '4.12'
     kafkaVersion = '1.1.1'
     mockitoVersion = '1.10.19'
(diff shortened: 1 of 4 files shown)


C0urante:
@Riduidel @criccomini @mtagle ping :)

Riduidel:
Sorry, I think I've tried your solution. Unfortunatly, as kafka connectors don't inherit the kafka connect configuration (I made that sure by looking at connector through debugger), and my schema registry is hosted on aiven - which requires https connection - with specific SSL certificates, well, it didn't work.
However, the solution seems good, and it deserves to be merged.

brunodomenici:
Hey guys, can you please release a version with this implementation? I've tested here with my env but I had to compile from master.
Nice job!
Thanks a lot!

PR #154, opened 2019-03-21
Support passing GCP credentials as JSON in config

ivanyu:
Support passing the GCP credentials via the configuration directive `credentials` as a non-escaped JSON blob.

This helps to run the Kafka connector when the environment can't be manipulated.

Changes:
- `credentials` configuration keys added. It has the priority over `keyfile` when provided.
- `GoogleCredentials` objects are now provided by `BigQuerySinkConfig.getCredentials`, which handles both `keyfile` and `credentials` configuration keys.
- The ability to use explicit JSON credentials added to the integration test suite:
  - support of `-c|--credentials` CLI parameter, `KCBQ_TEST_CREDENTIALS` environment variable, and `credentials` configuration key in `integrationtest.sh`; using them in all relevant pieces of config;
  - support of `kcbq_test_credentials` in `integrationTestPrep` Gradle task and passing it to `BucketClearer` and `TableClearer`.
- `README.md` updated.

### DIFF
--- a/kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/utils/BucketClearer.java
+++ b/kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/utils/BucketClearer.java
@@ -18,10 +18,10 @@ package com.wepay.kafka.connect.bigquery.it.utils;
  */
 
 
-import com.google.cloud.storage.Bucket;
-
+import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.storage.Storage;
 import com.wepay.kafka.connect.bigquery.GCSBuilder;
+import com.wepay.kafka.connect.bigquery.GoogleCredentialsReader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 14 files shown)


kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/utils/BucketClearer.java
```diff
@@ -50,7 +61,8 @@ public static void main(String[] args) {
 
   private static void usage() {
     System.err.println(
-        "usage: BucketClearer <key_file> <project_name> <bucket_name>"
+        "usage: BucketClearer <key_file> <credentials> <project_name> <bucket_name>\n" +
```
mtagle: I'm not a huge fan of this interface, but I'm willing to live with it because this is mostly just called from gradle anyway. However, I think here we should clarify that if key_file or credentials is empty, it's treated as non-existent. (This would also apply to TableClearer)

ivanyu: Thank you, @mtagle, I added this.

ivanyu:
Hi, I rebased on the latest `master`. Would you consider merging this?

ivanyu:
The CLA page says
image

criccomini:
@ivanyu , could you please try again on the CLA and rebase?

ivanyu:
@criccomini it seems that what I tried to achieve here was introduced in https://github.com/wepay/kafka-connect-bigquery/commit/2d49a94f94eb62477c65b966782564d008e3be64

However, this new `keySource` configuration is not documented in README, so I'll create a new PR.

PR #157, opened 2019-04-12
Replicate as much as possible Kafka Connect's `topics` config definition

ivanyu:
BigQuerySinkConfig has its own definition for configuration `topics`,
which overrides the one made by Kafka Connect.

The main drawback with this currently is that BigQuerySinkConfig doesn't
define configuration group, which is a problem with applications that
rely on configuration grouping (like UIs).

This commit adds to BigQuerySinkConfig' (re-)definition the group, the
position in the group, the width and display text; also it modifies the
documentation string. The values for these constants are taken from
Kafka Connect's code of 1.1.1 version (used currently by the project),
namely from https://github.com/apache/kafka/blob/1.1.1/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SinkConnectorConfig.java#L33

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
@@ -54,11 +54,16 @@ public class BigQuerySinkConfig extends AbstractConfig {
   private static final ConfigDef config;
   private static final Validator validator = new Validator();
 
+  // Values taken from https://github.com/apache/kafka/blob/1.1.1/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/SinkConnectorConfig.java#L33
   public static final String TOPICS_CONFIG =                     SinkConnector.TOPICS_CONFIG;
   private static final ConfigDef.Type TOPICS_TYPE =              ConfigDef.Type.LIST;
   private static final ConfigDef.Importance TOPICS_IMPORTANCE =  ConfigDef.Importance.HIGH;
+  private static final String TOPICS_GROUP =                     "Common";
+  private static final int TOPICS_ORDER_IN_GROUP =               4;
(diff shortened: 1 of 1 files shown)


C0urante:
@ivanyu would it be backwards compatible to just remove the `topics` config altogether? The Connect framework will still enforce the requirement that it or `topics.regex` be provided.

ivanyu:
@C0urante right, removing it altogether might be a better option. No other connector I know specify it explicitly, because as you said the Connect framework enforces it anyway. I will create a new PR for this.

ivanyu:
Sorry, I realized that this configuration defined explicitly is actually needed. Despite the Connect framework enforces its presence for validation etc, without it being defined it's impossible to read it in `BigQuerySinkConfig.getTopicsToDatasets`.
So, reopening the PR.

C0urante:
Hmmm... the `topics` config doesn't have a default value (both with the connector as it currently stands and with proposed changes in the PR). Doesn't this make it impossible to run the connector with the `topics.regex` config that Connect supports now?

C0urante:
I think an issue here is that when I wrote this initially I assumed the list of topics would be static and known at configuration time. With the addition of support for the `topics.regex` config that's no longer the case.

One approach could be to change the checks for overlap in `topic -> dataset` and `topic -> table` mappings from configuration time to runtime and cache mappings every time a new topic is encountered. The old checks could be retained if the `topics` config is used if it's a big concern that overlap be discovered ASAP instead of at runtime.

ivanyu:
> Doesn't this make it impossible to run the connector with the `topics.regex` config that Connect supports now?

It does. It's definitely worth changing. For example, we can override `open` in `BigQuerySinkTask` and check mappings there. The mappings are already done with regular expressions, so it should be possible to rework.

However, this may take some time so I'd propose to merge this PR anyway. Despite the introduction of `topics.regex` will make it redundant, it's fairly small and non-intrusive, but immediately useful (at least for people who show connector configs in UI, like in my case).

PR #159, opened 2019-05-29
Await termination on correct executor

anderseriksson:
We had problems with duplicate messages when restarting the task, so we think this is at least part of the problem.


mtagle:
oops, nice catch!

PR #164, opened 2019-06-20
Add prefix/folder related configuration.

bingqinzhou:
Add the prefix/folder configuration so that records from different connectors will be saved in different folders under the common GCS bucket.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -177,7 +177,7 @@ public class BigQuerySinkTask extends SinkTask {
           if (config.getList(config.ENABLE_BATCH_CONFIG).contains(record.topic())) {
             String gcsBlobName = record.topic() + "_" + uuid + "_" + Instant.now().toEpochMilli();
             String gcsFolderName = config.getString(config.GCS_FOLDER_NAME_CONFIG);
-            if (!gcsFolderName.equals("")) {
+            if (!"".equals(gcsFolderName)) {
               gcsBlobName = gcsFolderName + "/" + gcsBlobName;
             }
             tableWriterBuilder = new GCSBatchTableWriter.Builder(
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -175,7 +175,7 @@ public class BigQuerySinkTask extends SinkTask {
         if (!tableWriterBuilders.containsKey(table)) {
           TableWriterBuilder tableWriterBuilder;
           if (config.getList(config.ENABLE_BATCH_CONFIG).contains(record.topic())) {
-            String gcsBlobName = record.topic() + "_" + uuid + "_" + Instant.now().toEpochMilli();
+            String gcsBlobName = config.getString(config.GCS_FOLDER_NAME_CONFIG) + "/" + record.topic() + "_" + uuid + "_" + Instant.now().toEpochMilli();
             tableWriterBuilder = new GCSBatchTableWriter.Builder(
                 gcsToBQWriter,
                 table.getBaseTableId(),
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -177,7 +177,7 @@ public class BigQuerySinkTask extends SinkTask {
           if (config.getList(config.ENABLE_BATCH_CONFIG).contains(record.topic())) {
             String gcsBlobName = record.topic() + "_" + uuid + "_" + Instant.now().toEpochMilli();
             String gcsFolderName = config.getString(config.GCS_FOLDER_NAME_CONFIG);
-            if (!"".equals(gcsFolderName)) {
+            if (gcsFolderName != null && !"".equals(gcsFolderName)) {
               gcsBlobName = gcsFolderName + "/" + gcsBlobName;
             }
             tableWriterBuilder = new GCSBatchTableWriter.Builder(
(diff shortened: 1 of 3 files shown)


criccomini:
I don't recall--does subfolder get auto-created if it doesn't exist?

bingqinzhou:
> I don't recall--does subfolder get auto-created if it doesn't exist?

Reading about the following documentation and stack overflow discussions:

https://cloud.google.com/storage/docs/gsutil/addlhelp/HowSubdirectoriesWork
https://stackoverflow.com/questions/25214900/google-storage-api-list-storage-bucket-with-in-the-name 
https://stackoverflow.com/questions/38416598/how-to-create-an-empty-folder-on-google-storage-with-google-api

It feels like GCS doesn’t really have a physical conception about “folder", and when an object with a name like “folder_a/obj_1” is created, some visual tool makes it looks like that obj_1 is under folder_a.

criccomini:
Gotcha.

Another thought: this change is subtle different in prior behavior in that it will always at least have a `/` as the prefix. Again, I think this will work, but I'm not 100% sure.

Have you verified this change locally or via integration test? If not, I recommend at minimum verifying that things work locally.

bingqinzhou:
> Gotcha.
> 
> Another thought: this change is subtle different in prior behavior in that it will always at least have a `/` as the prefix. Again, I think this will work, but I'm not 100% sure.
> 
> Have you verified this change locally or via integration test? If not, I recommend at minimum verifying that things work locally.

Nice catch!

I feel that we shouldn't have a "/" prefix when folder config is not passed in. I updated the codes to not to include any prefix (as before) when the folder config is not passed in.

For integration tests, I updated the `integrationtest.sh` file to accept testing with the folder config, and tested that under both situations when the folder config is passed and not passed in, the testGCSLoad() passes.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -176,6 +176,10 @@ public void put(Collection<SinkRecord> records) {
           TableWriterBuilder tableWriterBuilder;
           if (config.getList(config.ENABLE_BATCH_CONFIG).contains(record.topic())) {
             String gcsBlobName = record.topic() + "_" + uuid + "_" + Instant.now().toEpochMilli();
+            String gcsFolderName = config.getString(config.GCS_FOLDER_NAME_CONFIG);
+            if (!gcsFolderName.equals("")) {
```
criccomini: For safety, `"".equals(gcsFolderName)`

bingqinzhou: Thanks for catching this! Updated accordingly.

criccomini: Actually, thinking on this more. My suggestion was not quite correct. I suspect config.getString() returns null if not set, and thus this if check would return true if it's null. Probably want != null

bingqinzhou: Thanks for pointing this out! The `gcsFolderName` config is set to be "" by default, but I added a `!= null` check to be safe :)

criccomini:
LGTM!

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -84,6 +84,14 @@
       "The name of the bucket in which gcs blobs used to batch load to BigQuery "
       + "should be located. Only relevant if enableBatchLoad is configured.";
 
+  public static final String GCS_FOLDER_NAME_CONFIG =                     "gcsFolderName";
+  private static final ConfigDef.Type GCS_FOLDER_NAME_TYPE =              ConfigDef.Type.STRING;
+  private static final Object GCS_FOLDER_NAME_DEFAULT =                   "";
+  private static final ConfigDef.Importance GCS_FOLDER_NAME_IMPORTANCE =  ConfigDef.Importance.MEDIUM;
+  private static final String GCS_FOLDER_NAME_DOC =
+          "The name of the folder under the bucket in which gcs blobs used to batch load to BigQuery "
```
mtagle: I think we should add a note that if it's unconfigured we'll upload into the root of the bucket.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -176,6 +176,10 @@ public void put(Collection<SinkRecord> records) {
           TableWriterBuilder tableWriterBuilder;
           if (config.getList(config.ENABLE_BATCH_CONFIG).contains(record.topic())) {
             String gcsBlobName = record.topic() + "_" + uuid + "_" + Instant.now().toEpochMilli();
+            String gcsFolderName = config.getString(config.GCS_FOLDER_NAME_CONFIG);
+            if (gcsFolderName != null && !"".equals(gcsFolderName)) {
```
mtagle: && !gcsFolderName.isEmpty() seems better to me.

kcbq-connector/test/integrationtest.sh
```diff
@@ -37,10 +37,11 @@ usage() {
        "[-p|--project <BigQuery project>]\n" \
        "[-d|--dataset <BigQuery project>]\n" \
        "[-b|--bucket <cloud Storage bucket>\n]" \
+       "[-f|--folder <cloud Storage folder under bucket>\n]" \
```
mtagle: is specifying the folder optional to run integration tests?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -176,6 +176,10 @@ public void put(Collection<SinkRecord> records) {
           TableWriterBuilder tableWriterBuilder;
           if (config.getList(config.ENABLE_BATCH_CONFIG).contains(record.topic())) {
             String gcsBlobName = record.topic() + "_" + uuid + "_" + Instant.now().toEpochMilli();
```
mtagle: maybe this is personal preference, but I feel like this would be more readable if we started with (maybe) having a folder and then adding the actual blob name onto it, rather than going in the other direction.

PR #165, opened 2019-06-23
Put git hash in non-release version numbers

wrp:
This commit will use git tags to generate version numbers in gradle.properties.  The current release method requires hand editing gradle.properties, and encourages building multiple distinct versions all with the name x.x.x-SNAPSHOT.  With this change, a git hash will be used instead of the static string SNAPSHOT, making it possible to determine the source of the build.   Until a tag is made on the mainline, this will cause ./gradlew to fail

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,23 +61,19 @@ $ mkdir bin/jar/ && tar -C bin/jar/ -xf kcbq-confluent/build/distributions/kcbq-
 ### Setting up smudge filter
 Rather than hand editing gradle.properties to change the version, it is relatively simple to
 set up a smudge filter which will automagically insert a version based on git tags.  To do so,
-add the following line to .gitattributes:
+run the following line:
 
-gradle.properties filter=id
-
-Then, create smudge and clean filters by executing:
-
+```bash
+gradle.properties filter=id > .gitattributes
 git config filter.id.clean 'git show HEAD:./gradle.properties'
 git config filter.id.smudge 'awk '"'"'$1 == "version" { "git describe --tags" | getline version; $2 = version; } 1'"'"' FS== OFS=='
+```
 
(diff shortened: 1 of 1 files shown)


wrp:
Changed to use the current version in gradle.properties (1.1.0-SNAPSHOT) rather than aborting in the absence of a valid tag.

criccomini:
:( gradlew is a standard file that's always committed as-is from gradle. I don't think it's recommended to mutate it at all.

https://docs.gradle.org/current/userguide/gradle_wrapper.html

wrp:
Then we should not use it.  Either we write our own script named gradlew,
or we write our own wrapper that invokes gradle wrapper, or we keep the
practice of releasing multiple indistinguishable versions named
*-x.x.x-SNAPSHOT.   I strongly oppose the last option.  There needs to be a
mechanism for distinguishing between developmental releases beyond
hand-editing the gradle.properties.  Also, it seems bad practice to have
the main versions tagged in stubby branches off the mainline rather than
have them as signed tags on the master.  I believe gradle.properties should
be a derived file rather than have it in the git repo.  If we build the
file in gradlew or in some other script is not particularly important.
Building it from gradlew seems most natural, though.

criccomini:
I think you're just battling the Java ecosystem.

https://stackoverflow.com/questions/5901378/what-exactly-is-a-maven-snapshot-and-why-do-we-need-it

SNAPSHOT is, again, a standard Maven release thingy. It bypasses dependency caching in build systems, among other things.

The value in diverging from the standard Java ecosystem that pretty much the whole world uses is minimal, IMO, compared to the cost of doing so.

I should not that we can certainly make the CI never build/release SNAPSHOT versions, but that's different than having developers default to SNAPSHOT on their local build.

PR #166, opened 2019-06-24
Refactor integration test

wrp:
This includes a lot of cleanup of the integration test, but the primary purpose is to make it run on linux.  The primary issues it addresses are differing behaviors of xargs and basename.  Currently, the script expects xargs to behave as if --no-run-if-empty was given (this is the default behavior on macos, but is not default in gnu xargs common on linux hosts), and it expects basename to take multiple arguments (this causes an error in gnu coreutils 8.26).  Also, I added some flags to keep the containers running if desired to enable further analysis after the test has completed.

### DIFF
--- a/kcbq-connector/test/integrationtest.sh
+++ b/kcbq-connector/test/integrationtest.sh
@@ -94,8 +94,9 @@ init() {
 }
 
 parse_args() {
+	folder=kcbq-test  # assign default
 	# Process command line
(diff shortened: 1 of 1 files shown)


stoynov96:
If this is just a cleanup now, it may be a good idea to update the PR title

mtagle:
Going to close this in favor of the other PRs.

PR #168, opened 2019-07-03
Support Java 11

ivanyu:
This commit upgrade Gradle used in the project to the latest available
version 5.5 to support Java 11.
It also switches from (now deprecated) FindBugs plugin in SpotBugs for
the same reason.
`gradle/wrapper` directory and `gradlew.{bat|sh}` are re-generated
automatically by Gradle.


ivanyu:
Sorry, missed this issue with Javadoc, fixing it.

ivanyu:
Would you consider merging this?

mtagle:
Yes! I just want to verify that this will still continue to work with older versions of java?

ivanyu:
@mtagle it will. It's checked on Travis.
Also, I found that `sourceCompatibility` hasn't been set correctly. I fixed this in a new commit and now it should always target compatibility with Java 8 despite the JDK version being used for building.

PR #170, opened 2019-07-11
Fix gcs load retry logic

mtagle:
wrote a bunch of tests for GCSToBQWriter

fixed retry logic for GCSToBQWriter:
- catch the error thrown by GCS if the upload is unsuccessful (and thus actually retry).
- keep track of whether we actually succeeded in our retry attempts or not/ don't suggest we succeeded in a log message if we did not.
- stop retrying once we succeed in a retry attempt.

it's possible that this will fix the issues we've been seeing with dropped rows when going through GCS loading.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
@@ -75,9 +75,9 @@ public class GCSBatchTableWriter implements Runnable {
     try {
       writer.writeRows(rows, tableId, bucketName, blobName);
     } catch (ConnectException ex) {
-      throw new ConnectException("Failed to write rows to GCS", ex);
+      throw new ConnectException(String.format("Failed to write rows to GCS for table %s", tableId.getTable()), ex);
     } catch (InterruptedException ex) {
-      throw new ConnectException("Thread interrupted while batch writing to GCS.", ex);
+      throw new ConnectException(String.format("Thread interrupted while batch writing to GCS for table %s", tableId.getTable()), ex);
     }
   }
 
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
@@ -75,9 +75,9 @@ public class GCSBatchTableWriter implements Runnable {
     try {
       writer.writeRows(rows, tableId, bucketName, blobName);
     } catch (ConnectException ex) {
-      throw new ConnectException(String.format("Failed to write rows to GCS for table %s", tableId.getTable()), ex);
+      throw new ConnectException("Failed to write rows to GCS", ex);
     } catch (InterruptedException ex) {
-      throw new ConnectException(String.format("Thread interrupted while batch writing to GCS for table %s", tableId.getTable()), ex);
+      throw new ConnectException("Thread interrupted while batch writing", ex);
     }
   }
 
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
@@ -74,8 +74,10 @@ public class GCSBatchTableWriter implements Runnable {
   public void run() {
     try {
       writer.writeRows(rows, tableId, bucketName, blobName);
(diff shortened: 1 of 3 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
```diff
@@ -74,8 +74,10 @@ private GCSBatchTableWriter(List<RowToInsert> rows,
   public void run() {
     try {
       writer.writeRows(rows, tableId, bucketName, blobName);
-    } catch (ConnectException | InterruptedException ex) {
-      throw new ConnectException("Thread interrupted while batch writing to BigQuery.", ex);
+    } catch (ConnectException ex) {
+      throw new ConnectException("Failed to write rows to GCS", ex);
+    } catch (InterruptedException ex) {
+      throw new ConnectException("Thread interrupted while batch writing to GCS.", ex);
```
wrp: Is it easy at this point to indicate the name of the connector in the log?  Or some identifying mark so we can know which connector is generating the error.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
```diff
@@ -108,23 +109,28 @@ public void writeRows(List<RowToInsert> rows,
...
       }
       retryCount++;
-    } while (exceptionsOccurred && (retryCount < retries));
+    } while (!success && (retryCount < retries));
+
+    if (success) {
+      logger.info("Batch loaded {} rows", rows.size());
+    }
+    else {
+      throw new ConnectException(String.format("Failed to load %d rows into GCS within %d attempts.", rows.size(), retries));
```
wrp: Similarly here and in line 128 (logger.info("Batch loaded...).  It would be nice to be able to identify the connector that is generating the error.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
mtagle: hm, I'll look into it.

criccomini: @mtagle you might look to see if Kafka connect now includes connector info in the log context ([MDC](https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/MDC.html)). If so, SREs just need to include a new pre-defined variable.

Yup. https://issues.apache.org/jira/browse/KAFKA-3816 .. I recommend just using this once we get to 2.3, since it looks like this is when it was added.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
```diff
@@ -108,23 +109,28 @@ public void writeRows(List<RowToInsert> rows,
...
         uploadRowsToGcs(rows, blobInfo);
-      } catch (ConnectException ce) {
-        exceptionsOccurred = true;
+        success = true;
+      } catch (StorageException se) {
         logger.warn("Exceptions occurred for table {}, attempting retry", tableId.getTable());
       }
       retryCount++;
-    } while (exceptionsOccurred && (retryCount < retries));
+    } while (!success && (retryCount < retries));
```
criccomini: Possible to move this to a normal while() rather than do/while as a best-practices thing?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
```diff
@@ -108,23 +109,28 @@ public void writeRows(List<RowToInsert> rows,
...
+
+    if (success) {
+      logger.info("Batch loaded {} rows", rows.size());
+    }
+    else {
```
criccomini: Nit: Not sure , but :

```
}
else {
```

Seems not to style.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
criccomini: MDC. See above.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
mtagle: so I can't give the connector name but I can give the name of the table it's attempting to get data into. That is hopefully helpful?

criccomini:
LGTM!

PR #184, opened 2019-08-22
Resolve table name dynamically

CodingParsley:
Currently the table lookup map in BigQuery connector only include topics passed in to the connector config file. To dynamically route messages to different topics and thus different tables, big query connector should be able to evaluate the topic names embedded in messages automatically and include them in the look up map, so that the messages can be routed to different BigQuery tables.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableUpdater.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableUpdater.java
@@ -1,0 +1,92 @@
+package com.wepay.kafka.connect.bigquery.utils;
+
+import com.google.cloud.bigquery.TableId;
+import com.wepay.kafka.connect.bigquery.config.BigQuerySinkConfig;
+import org.apache.kafka.common.config.ConfigException;
+
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.wepay.kafka.connect.bigquery.utils.TopicToTableResolver.getTopicToTableSingleMatch;
+
+public class TopicToTableUpdater {
+  public static String getTopicToDateset (
+          List<Map.Entry<Pattern, String>> patterns,
+          String topicName,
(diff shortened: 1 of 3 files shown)


apoorvmittal10:
@criccomini Please if you can help closing this PR.

whynick1:
> Currently the table lookup map in BigQuery connector only include topics passed in to the connector config file.
That's true. If you want to add a new topic, you need to update connector config file and restart it.
> To dynamically route messages to different topics and thus different tables, big query connector should be able to evaluate the topic names embedded in messages automatically and include them in the look up map, so that the messages can be routed to different BigQuery tables.
KCBQ only subscribes to topic in its config file, which means it can always resolve table name for sink record with current table lookup map. Can you explain why we need `Resolve table name dynamically`?

criccomini:
My suspicion is that KC can now support dynamic topic subscription (via regex maybe?). If so, a patch like this would be needed so that, as new topics are created that match the consumer regex, they get properly routed to BQ tables.

@CodingParsley can you confirm? cc @apoorvmittal10

criccomini:
Probably related to this? https://issues.apache.org/jira/browse/KAFKA-3073

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableUpdater.java
```diff
@@ -0,0 +1,92 @@
...
+    return match;
+  }
+
+  /**
+   * Return a Map detailing which BigQuery table each topic should write to.
+   *
+   * @param config Config that contains properties used to generate the map
+   * @return A Map associating Kafka topic names to BigQuery table names.
+   */
+  public static Map<String, TableId> updateTopicToTable(
```
bingqinzhou: You probably don't need to return a map here since the passed in `topicToTable` map will be modified.

apoorvmittal10:
Refactored code and streamlined code with other PRs as well. I need to open separate PR #189 for same as @CodingParsley is not currently part of ConfluentInc.

@CodingParsley Please if we can close this PR.

@bingqinzhou Apologies for the trouble.

PR #185, opened 2019-08-23
custom SchemaRetriever 

CodingParsley:


whynick1:
@CodingParsley Can you explain why you want read schema from record rather than from `schemaCache`?

PR #181, opened 2019-08-14
Add Support for Schemaless Record

CodingParsley:
What
----
Currently the BigQueryRecordConverter throws an error when schemaless JSON data is consumed. JsonConverter converts JSON to map before it arrives at the BigQuery connector. Therefore, instead of throwing null pointer error, the BigQueryRecordConverter should deal with schemaless data type by converting it to legitimate format.

Should solve this issue -> https://github.com/wepay/kafka-connect-bigquery/issues/174 

@C0urante 
Review is greatly appreciated!

PR #188, opened 2019-09-17
Add Support for Schemaless Record

apoorvmittal10:
* Currently the BigQueryRecordConverter throws an error when schemaless JSON data is consumed.
* JsonConverter converts JSON to map before it arrives at the BigQuery connector. Therefore, instead of throwing null pointer error, the BigQueryRecordConverter should deal with schemaless data type by converting it to legitimate format.
* Should also fix https://github.com/wepay/kafka-connect-bigquery/issues/174

This is the PR against #181 . As ConfluentInc has signed corporate CLA with WePay and @CodingParsley is currently not part of ConfluentInc hence raised separate PR. Please if we can merge same and close #181 .

@criccomini @mtagle


apoorvmittal10:
@mtagle Gentle reminder.

PR #190, opened 2019-09-23
WIP: Add keyFileType config to support keyFile as string

ilanjiR:
With this PR, we add a keyFileType config which defaults to "FILE". When this config is set to "JSON", the connector will treat the keyFile config as a string representation of the Google Credentials object rather than as the path to the Google Credentials file.


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -132,6 +132,13 @@
   private static final String KEYFILE_DOC =
       "The file containing a JSON key with BigQuery service account credentials";
 
+  public static final String KEYFILE_TYPE_CONFIG =                "keyfiletype";
+  private static final ConfigDef.Type KEYFILE_TYPE_TYPE =         ConfigDef.Type.STRING;
+  public static final String KEYFILE_TYPE_DEFAULT =               null;
```
mageshn: I think the default should be `FILE`

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -239,6 +246,12 @@
             KEYFILE_DEFAULT,
             KEYFILE_IMPORTANCE,
             KEYFILE_DOC
+        ).define(
+            KEYFILE_TYPE_CONFIG,
+            KEYFILE_TYPE_TYPE,
+            KEYFILE_TYPE_DEFAULT,
```
mageshn: may include a validator that limits the values to `FILE` and `JSON`

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/GCSBuilder.java
```diff
@@ -36,39 +37,53 @@
   private static final Logger logger = LoggerFactory.getLogger(GCSBuilder.class);
 
   private final String projectName;
-  private String keyFileName;
+  private String keyFile;
+  private String keyFileType;
 
-  public GCSBuilder(String projectName) {
+  public GCSBuilder(String projectName, String keyFileType) {
```
mageshn: we should add `setKeyFileType` rather than adding an arg to the constructor to preserve the existing pattern.

PR #192, opened 2019-10-04
Log topic and table name when a row fails to be inserted into BQ

bingqinzhou:


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -182,11 +182,13 @@ private boolean isPartialFailure(List<InsertAllRequest.RowToInsert> rows,
...
   private List<InsertAllRequest.RowToInsert> getFailedRows(List<InsertAllRequest.RowToInsert> rows,
-                                                           Set<Long> failRowsSet) {
+                                                           Set<Long> failRowsSet,
+                                                           String topic,
+                                                           PartitionedTableId table) {
     List<InsertAllRequest.RowToInsert> failRows = new ArrayList<>();
     for (int index = 0; index < rows.size(); index++) {
       if (failRowsSet.contains((long)index)) {
-        logger.debug("[row with index {}]: failed", index);
+        logger.debug("Row with index {} from topic {} failed to be written to table {}.", index, topic, table.getFullTableName());
```
mtagle: I feel like this is a good time to maybe move this log message outside of the for loop so we don't get a big stack of errors, but instead one error message that contains all the indexes that failed.

bingqinzhou: > I feel like this is a good time to maybe move this log message outside of the for loop so we don't get a big stack of errors, but instead one error message that contains all the indexes that failed.

But how should we log the tables and topics of a batch of rows in one logging message in this case?

PR #198, opened 2019-10-18
Log errors encountered when writing batches

wicknicks:
We run into the following error with the BigQuery connector

```
org.apache.kafka.connect.errors.ConnectException: Attempted to reduce batch size below 1.
    at com.wepay.kafka.connect.bigquery.write.batch.TableWriter.getNewBatchSize(TableWriter.java:106)
```
It is not clear if the problem was that the original set of records was too few, or there were problems writing batches to BigQuery and the connector retried because it misinterpreted the error codes [here](https://github.com/wepay/kafka-connect-bigquery/blob/1.2.0/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java#L83-L87).  

Adding this log message should give us some more context on what went wrong. 

Signed-off-by: Arjun Satish <arjun@confluent.io>

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
@@ -81,6 +81,7 @@ public class TableWriter implements Runnable {
           currentIndex += currentBatchSize;
           successCount++;
         } catch (BigQueryException err) {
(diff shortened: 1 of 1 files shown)


wicknicks:
@rhauch @C0urante can you take a first pass at this PR? thanks!

wicknicks:
@mtagle @criccomini could you please take a pass at this. thanks!

mtagle:
I'm always in favor of helpful logging!

Issue #86, opened 2017-05-08
Using the connector with Json-based topic


ofirsharony:
Hi guys,

I've successfully used the connector to stream data from an avro-based topic to BigQuery.
Does the connector support streaming Json-based topics as well?

Thanks.

criccomini:
I believe so, but I've not given it a try personally. I think @C0urante set things up so that kcbq-confluent is a separate package from kcbq-connector (which has the core connector stuff).

That said, I think you'll have to implement a [SchemaRetriever](https://github.com/wepay/kafka-connect-bigquery/blob/master/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java) to make the connector work. With the Confluent implementation, the schema is fetched from the Avro schema registry. Without a schema registry, you'll have to get the proper schema for a given table from somewhere (or you'll have to manually create the tables or something).

@C0urante and @mtagle might have more thoughts on this.

Please post your findings. I think you might be the first to use JSON with this. :) Would be good to get feedback.

C0urante:
Believe @criccomini has it right. Theoretically (I have to qualify this since I haven't actually seen it tested yet and things may have changed since the last time I touched this project), all you'd have to worry about would be A) ensuring the necessary tables were created somehow (either by hand or via the SchemaRetriever API) and B) making sure that the JSON data flowing into the connector matched the format of the table(s) it would be flowing into.

criccomini:
@ofirsharony any update here?

ofirsharony:
@criccomini Not at the moment. 
I've decided to work only with Avro-based topics for now, saving on storage and using its schema evolution.

natb1:
I just looked in to this. I don't think it's as simple creating the table manually and configuring the connector to disable schemas.

i.e. this following config:
```
https://github.com/wepay/kafka-connect-bigquery/blob/master/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java#L69
```
resulted in this exception for me:
```
at com.wepay.kafka.connect.bigquery.BigQuerySinkTask.put(BigQuerySinkTask.java:165)
at com.wepay.kafka.connect.bigquery.BigQuerySinkTask.getRecordRow(BigQuerySinkTask.java:143)
at com.wepay.kafka.connect.bigquery.convert.BigQueryRecordConverter.convertRecord(BigQueryRecordConverter.java:51)
[...]

PR #199, opened 2019-10-21
Initial commit/foundation to support connect record schema retriever

apoorvmittal10:
Support for json based topics has already been fixed under https://github.com/wepay/kafka-connect-bigquery/issues/174 though `autoUpdateSchemas` and `autoCreateTables` cannot be supported unless we have different schema retriever (other than using `schemaRegistry`).

To support `ConnectRecord` based schema retriever we need to use `ConnectRecord.valueSchema` hence this PR builds the foundation to support same.

Another issue which highlights the requirement for this behaviour https://github.com/wepay/kafka-connect-bigquery/issues/86.

@criccomini @mtagle @C0urante Please review.

In subsequent PR I shall go forward and implement connect record schema retriever.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -31,7 +31,6 @@ import com.wepay.kafka.connect.bigquery.config.BigQuerySinkTaskConfig;
 import com.wepay.kafka.connect.bigquery.convert.RecordConverter;
 import com.wepay.kafka.connect.bigquery.convert.SchemaConverter;
 import com.wepay.kafka.connect.bigquery.exception.SinkConfigConnectException;
-import com.wepay.kafka.connect.bigquery.utils.FieldNameSanitizer;
 import com.wepay.kafka.connect.bigquery.utils.PartitionedTableId;
 import com.wepay.kafka.connect.bigquery.utils.TopicToTableResolver;
 import com.wepay.kafka.connect.bigquery.utils.Version;
@@ -72,6 +71,7 @@ public class BigQuerySinkTask extends SinkTask {
   private static final Logger logger = LoggerFactory.getLogger(BigQuerySinkTask.class);
 
   private SchemaRetriever schemaRetriever;
+  private BigQuery bigQuery;
   private BigQueryWriter bigQueryWriter;
   private GCSToBQWriter gcsToBQWriter;
   private BigQuerySinkTaskConfig config;
@@ -83,7 +83,7 @@ public class BigQuerySinkTask extends SinkTask {
 
   private KCBQThreadPoolExecutor executor;
   private static final int EXECUTOR_SHUTDOWN_TIMEOUT_SEC = 30;
-  
+
   private final BigQuery testBigQuery;
   private final Storage testGcs;
 
@@ -149,22 +149,6 @@ public class BigQuerySinkTask extends SinkTask {
     return builder.build();
   }
 
-  private RowToInsert getRecordRow(SinkRecord record) {
-    Map<String,Object> convertedRecord = recordConverter.convertRecord(record);
-    if (config.getBoolean(config.SANITIZE_FIELD_NAME_CONFIG)) {
-      convertedRecord = FieldNameSanitizer.replaceInvalidKeys(convertedRecord);
-    }
-
-    return RowToInsert.of(getRowId(record), convertedRecord);
-  }
-
-  private String getRowId(SinkRecord record) {
-    return String.format("%s-%d-%d",
-        record.topic(),
-        record.kafkaPartition(),
-        record.kafkaOffset());
-  }
-
   @Override
   public void put(Collection<SinkRecord> records) {
     logger.info("Putting {} records in the sink.", records.size());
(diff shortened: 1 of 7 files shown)


apoorvmittal10:
@mtagle @criccomini Please if you could review same.

apoorvmittal10:
>I sort of understand why you're trying to move back translating a sinkRecord into a InsertRecord for as long as possible, but I feel like getting the schema should still be something handled by the schemaRetriever, and I'm unclear how this refactoring is going to interact with whatever changes you are going to make there. Is there a reason you want to split this change into multiple PRs versus just one that clarifies how all of this is going to work together?

@mtagle `AdaptiveBigQueryWriter.java` handles dynamic update of schema in case error is thrown while writing to bigquery. `AdaptiveBigQueryWriter.java` was earlier taking input as processed rows to insert `InsertAllRequest.RowToInsert`. This will work in case when schema is retrieved from `schema registry` but when dealing with formats other than `avro` where `schema registry` is unaware of schema we need different different `schema retriever`. As connect records hold `value schema` hence the idea is to pull schema from `ConnectRecord`.  That is the reason this refactoring is needed. Please let me know if in case there could be any other solution too.

The reason this PR has been split in two because this PR laid the foundation for design changes and can be independently merged. The other PR will focus entirely on new connect record based schema retriever. Also the changes would have been significant and would made it hard to review hence I have split them in two.

mtagle:
Are you planning on modifying how SchemaRetriever works at all, or are you planning on having you code bypass it entirely? This sort of splitting leads me to believe you are planning on the latter but I don't think that's a good approach from a design sanity standpoint. If you are planning on modifying how SchemaRetriever works, I think this PR would make more sense with that change in that without.

I guess it's hard for me to say because I haven't seen the other half of this PR, but I'm not sure it's easier to review these changes because there are less of them when I don't understand how they are going to fit into the larger change.

whynick1:
> This will work in case when schema is retrieved from schema registry but when dealing with formats other than avro where schema registry is unaware of schema we need different different schema retriever. As connect records hold value schema hence the idea is to pull schema from ConnectRecord.

For Avro, we can easily retrieve schema from schema registry. 
For formats other than Avro like json, I am not sure how you manages the schema (another endpoint similar to schema registry, or something?). You should retrieve schema by implementing your own  `SchemaRetriever` rather than read from `ConnectRecord.valueSchema` (it could be `null`).
```
public interface SchemaRetriever {
  public void configure(Map<String, String> properties);
  public Schema retrieveSchema(TableId table, String topic);
  public void setLastSeenSchema(TableId table, String topic, Schema schema);
}
```
@apoorvmittal10

whynick1:
@apoorvmittal10 I take it back "You should retrieve schema by implementing your own SchemaRetriever". That's not the only options.

Talked with @mtagle and @criccomini . Here is a brief summary.
To support create/update BigQuery table schema with non-Avro Kafka message, we propose three options.
1. Provide a hard coded schema for each top, and create BigQuery table at start up
2. Extract schema from self-description payload. When first see a Kafka message of given top, retrieve schema by parsing its value, and use it to create BigQuery table. For any "schema change" that cause BigQuery update failure, we will repeat this step to update BQ schema.
3. Retrieve schema by implementing your own SchemaRetriever (not rely on `ConnectRecord`)

Base on your PR, it seems you try to implement the *2nd* approach. If so, some questions for you:
1) What you expect `ConnectRecord.valueSchema` to be for non-Avro Kafka message? Personally,   never tried to look at it. Could it be null? 
2) Why not use `ConnectRecord.value` or `RowToInsert` as our plan is to extract schema from `value`?

apoorvmittal10:
> For Avro, we can easily retrieve schema from schema registry.
For formats other than Avro like json, I am not sure how you manages the schema (another endpoint similar to schema registry, or something?). You should retrieve schema by implementing your own SchemaRetriever rather than read from ConnectRecord.valueSchema (it could be null).
```java public interface SchemaRetriever {
  public void configure(Map<String, String> properties);
  public Schema retrieveSchema(TableId table, String topic);
  public void setLastSeenSchema(TableId table, String topic, Schema schema);
}
```

@whynick1 Exactly this I am approaching for. Implement schema retriever that shall fetch `Schema` from `ConnectRecord` and update accordingly. 

>What you expect ConnectRecord.valueSchema to be for non-Avro Kafka message? Personally, never tried to look at it. Could it be null?

It shall be following: `https://kafka.apache.org/20/javadoc/org/apache/kafka/connect/data/Schema.html`. If used with transformation then it cannot be null.

>Why not use ConnectRecord.value or RowToInsert as our plan is to extract schema from value?

`ConnectRecord` does have `valueSchema` in itself. One shall use a transformation to fill `valueSchema`.

> I guess it's hard for me to say because I haven't seen the other half of this PR, but I'm not sure it's easier to review these changes because there are less of them when I don't understand how they are going to fit into the larger change.

Providing link to complete changes detail (commit) in my repo will work for complete view or I shall add changes here itself?

whynick1:
> ConnectRecord does have valueSchema in itself. One shall use a transformation to fill valueSchema.
Could you explain where and how you plan to do transformation to fill valueSchema? Is that in your 2nd half PR?
> Providing link to complete changes detail (commit) in my repo will work for complete view or I shall add changes here itself?
I will prefer to add the change here (with different commits).

apoorvmittal10:
>Could you explain where and how you plan to do transformation to fill valueSchema? Is that in your 2nd half PR?

Yes

> I will prefer to add the change here (with different commits).

Sure, I shall add them here. Currently occupied with other deliverable hence shall upload other patch in couple of days.

mtagle:
@apoorvmittal10 Just wanted to check in to see if this PR is still being worked on?

apoorvmittal10:
> @apoorvmittal10 Just wanted to check in to see if this PR is still being worked on?

@mtagle I am closing the PR for now as I am not actively working on same. I shall open an issue for supporting dynamic schema update for other data formats i.e. json and would discuss the approach regarding same prior proceeding with code changes.
Please let me know if you think differently.

PR #215, opened 2019-11-22
Use batch delete operation when cleaning up blob ids

farmdawgnation:
When we have a high number of blobs to delete from GCS (e.g. 10k) we're seeing the BQ Connector's batch load feature take upward of a half hour to delete all the blobs. Even with as few as 300 we can see delays of up to 60s.

This change significantly reduces the overhead involved by using the batch delete storage API to delete the blobs. In early testing I'm able to remove 500+ blobs in 3s, which is a significant improvement. This code preserves the same behavior previously with respect to the `deletableBlobIds` variable.


farmdawgnation:
Just want to add here that we're running this in prod now and have confirmed the improvement on production load. Delete operations that were taking 40 minutes are now taking 8 seconds.

farmdawgnation:
@bingqinzhou — Tagging you since you seem to be running releases lately. I think this is ready for code review whenever you can spare a moment. Thanks in advance!

mtagle:
This is great! I didn't realize there was a faster option for blob deletion. Thank you!

Issue #175, opened 2019-07-16
Support topic with multiple schemas


jaroslawZawila:
I am trying to use this connector with the topic with multiple schemas. The problem is the subject name for schema is `topic-record`  where this code is hard coded to `topic-value`. Are you going to support that kind of scenarios?

jaroslawZawila:
@mtagle If I would make the changes to add support for above , are you going to accept it?

mkubala:
I'm also ready to put some time & effort and prepare a pull request which would add support for multi schema topics. 
It would be great to know that you @mtagle @criccomini are not going to reject the whole idea.

mtagle:
I have to admit I'm a little unclear about what you mean by "a topic with multiple schemas". I'm particularly unclear on how a topic with multiple schemas would successfully end up in a table in BQ with (presumably) a single fixed schema. Could you elaborate on this? Maybe some details aobut your use-case would be helpful as well.

mkubala:
In our use case we are using an event sourcing approach where a single topic contains many different entities. We would like to store those entities in a dedicated BigQuery tables (one table for each event type) for further data ingestion / BI.

This approach is already supported by Kafka and Schema Registry (see https://www.confluent.io/blog/put-several-event-types-kafka-topic/).

We would add two new configuration options (names can be changed, this is just only draft):
- `recordNamesToTables` - A list of comma-delimited strings representing mapping between record names and table names (just like `topicsToTables` but for records). Default: empty list.
- `multipleSchemaTopics` - A boolean feature switch. Default: false.

Table names would contain mapped topic and record name (separated by `_`).

What do you think about the whole idea?

mtagle:
Thank you for the clarification, this makes a lot more sense to me now!

As the schema-registry already supports this, I'm in support of adding this feature to KCBQ. We'd be happy to review and merge your changes if you wanted to implement this.

mkubala:
Great! Expect a PR soon :)

PR #219, opened 2019-12-03
Add support for topics with multiple schemas (#175)

mkubala:
This contribution makes kafka-connect-bigquery compatible with topics that carry different types of messages, especially when Kafka is used for an event sourcing.

This approach is already supported by Kafka and Schema Registry (see https://www.confluent.io/blog/put-several-event-types-kafka-topic/).

What's been changed:
* Added support for topics with multiple record types (schemas) by adding support for TopicAndRecordName strategy,
* Added support for creating missing tables on the fly (no need to restart connector whenever new record type is being published to the multi schema topic),
* Moved `autoCreateTables` setting from `BigQuerySinkConnectorConfig` to `BigQuerySinkConfig`,
* After that `BigQuerySinkConnectorConfig` become no longer needed so I removed it,
* Added support for handling empty records (for both, SinkRecord and Schema converters) - for event envelopes with an empty payload.
* Cleaned code from unused methods and fields in various places


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConnectorConfig.java
```diff
@@ -1,82 +0,0 @@
...
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import java.util.Map;
-
-/**
- * Class for connector-specific configuration properties.
- */
-public class BigQuerySinkConnectorConfig extends BigQuerySinkConfig {
```
C0urante: Good night, sweet prince 😞

mkubala:
I'm going to add yet another improvement - when connector is running in a multi schema mode, we should be able to match dataset based on the topic and record name. This way we will be able to aggregate tables by their logical meaning (e.g. payment_events dataset with tables for successful and failed payments, order_events containing tables for orders' state changes etc.).

I decided to do it in a separate PR because this one is in my opinion already too large (sorry for that), but it would be great if you could include both of them in one release (or make subsequent release right after another PR will be approved and merged, which is what I really hope for 😉 ).

mkubala:
I forgot that I'm working on a forked repo and my feature branch is absent here, so I'm not able to expose a PR that's based on top of this particular branch. Looks like I have to add another commit to this PR.

jaroslawZawila:
What is going on with this PR?

mkubala:
I see that in the meantime a couple of other PRs have been merged and because this PR is quite big, conflicts appear. I'm going to fix them, hoping this PR will be reviewed soon :) cc  @mtagle @criccomini @C0urante

criccomini:
Hey all, just saw this. Sorry, lemme follow up with @bingqinzhou

criccomini:
This is a pretty big PR. @mkubala can you break it up?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -131,11 +132,23 @@ private PartitionedTableId getRecordTable(SinkRecord record) {
...
     // constructed when connector starts hence new topic configuration needed connector to restart.
     // Dynamic update shall not require connector restart and shall compute table id in runtime.
-    if (!topicsToBaseTableIds.containsKey(record.topic())) {
-      TopicToTableResolver.updateTopicToTable(config, record.topic(), topicsToBaseTableIds);
+    TopicAndRecordName topicAndRecordName = new TopicAndRecordName(record.topic(), record.valueSchema().name());
+    if (!topicsToBaseTableIds.containsKey(topicAndRecordName)) {
+      TopicToTableResolver.updateTopicToTable(config, topicAndRecordName, topicsToBaseTableIds);
+      TableId tableId = topicsToBaseTableIds.get(topicAndRecordName);
+      if (bigQuery.getTable(tableId) == null) {
+        boolean createMissingTables = config.getBoolean(BigQuerySinkConfig.TABLE_CREATE_CONFIG);
```
bingqinzhou: Can you please change the boolean name to be `autoCreateTables` for consistency?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -131,11 +132,23 @@ private PartitionedTableId getRecordTable(SinkRecord record) {
...
+        if (createMissingTables) {
+          logger.info("Table {} does not exist. Attempting to create.", tableId.getTable());
+          schemaManager.createTable(tableId, topicAndRecordName);
+        } else {
+          throw new ConnectException(
```
bingqinzhou: Do we need to specifically throw an exception here? Since if the table doesn't exist, bigquery will throw a corresponding exception while trying to insert records to that table.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -608,6 +706,23 @@ private void verifyBucketSpecified() throws ConfigException {
...
+    Class<?> schemaRetriever = getClass(BigQuerySinkConfig.SCHEMA_RETRIEVER_CONFIG);
+
+    boolean autoCreateTables = getBoolean(TABLE_CREATE_CONFIG);
+    if (autoCreateTables && schemaRetriever == null) {
+      throw new ConfigException(
+          "Cannot specify automatic table creation without a schema retriever"
+      );
+    }
+
+    if (schemaRetriever == null) {
```
bingqinzhou: Can you please remove this check since it's not necessary and confusing?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -131,11 +132,23 @@ private PartitionedTableId getRecordTable(SinkRecord record) {
...
-      TopicToTableResolver.updateTopicToTable(config, record.topic(), topicsToBaseTableIds);
+    TopicAndRecordName topicAndRecordName = new TopicAndRecordName(record.topic(), record.valueSchema().name());
+    if (!topicsToBaseTableIds.containsKey(topicAndRecordName)) {
+      TopicToTableResolver.updateTopicToTable(config, topicAndRecordName, topicsToBaseTableIds);
+      TableId tableId = topicsToBaseTableIds.get(topicAndRecordName);
+      if (bigQuery.getTable(tableId) == null) {
+        boolean createMissingTables = config.getBoolean(BigQuerySinkConfig.TABLE_CREATE_CONFIG);
+        if (createMissingTables) {
+          logger.info("Table {} does not exist. Attempting to create.", tableId.getTable());
+          schemaManager.createTable(tableId, topicAndRecordName);
```
mtagle: this feels out of scope for this PR. Is there a pr-relevant reason you can't just create the necessary tables on start up like the connector does at the moment?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -100,12 +110,29 @@
   private static final ConfigDef.Type TOPICS_TO_TABLES_TYPE =              ConfigDef.Type.LIST;
   private static final ConfigDef.Importance TOPICS_TO_TABLES_IMPORTANCE =
       ConfigDef.Importance.MEDIUM;
-  public static final Object TOPICS_TO_TABLES_DEFAULT =                    null;
```
mtagle: these sorts of changes feel unnecessary and not related to this PR (and, conceptually, I think having the default for a particular config be public is fine).

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConnectorConfig.java
mtagle: What exactly is the reason for removing this?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java
```diff
@@ -178,12 +178,17 @@ private Object convertObject(Object kafkaConnectObject, Schema kafkaConnectSchem
...
     Struct kafkaConnectStruct = (Struct) kafkaConnectObject;
     for (Field kafkaConnectField : kafkaConnectSchemaFields) {
-      Object bigQueryObject = convertObject(
-          kafkaConnectStruct.get(kafkaConnectField.name()),
-          kafkaConnectField.schema()
-      );
-      if (bigQueryObject != null) {
-        bigQueryRecord.put(kafkaConnectField.name(), bigQueryObject);
+      // ignore empty structures
+      boolean isEmptyStruct = kafkaConnectField.schema().type() == Schema.Type.STRUCT &&
```
mtagle: I'm a little confused about what's going on here or how it relates to the PR.

criccomini:
Echoing sentiment here:

1. PR is too big. Please split it up.
2. We will merge @bingqinzhou's PR. Her approach is roughly similar to yours. Please rebase with her changes once her PR is merged in (planning tomorrow AM).

Sorry for delay on this. Thanks for contributing--it's appreciated. :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
mkubala: sure

kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
```diff
@@ -16,22 +18,24 @@
...
    * @param properties The configuration settings of the connector.
    */
-  public void configure(Map<String, String> properties);
+  void configure(Map<String, String> properties);
 
   /**
    * Retrieve the most current schema for the given topic.
    * @param table The table that will be created.
-   * @param topic The topic to retrieve a schema for.
+   * @param topicAndRecordName The topic and an optional record name to retrieve a schema for.
```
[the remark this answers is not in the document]
mkubala: Record name is optional. I did not want to end up with methods that have too many arguments and overloads. 

It's been a while since I commercially worked with Java (I use Scala on my daily basis) and AFAICS passing parameters of type `Optional<T>` is not recommended (as one of the Java core team members, it's not aimed to be a counterpart of `Maybe` monad from Haskell or Scala's `Option`) which is reflected in my IDE's default warnings.

Passing nulls in place of a newly added optional parameters does not look good to me, hence plain `String` has been replaced with a dedicated type with an optional record type (which returns an `Optional<String>` and enforces its API users to handle the case with record type being absent. And that's the use case recommended by Java team members and authorities). 

Another motivation was to improve type safety, make method signatures more clear and avoid having some _util_ class with a static method that creates a subject from two strings.

kcbq-confluent/src/main/java/com/wepay/kafka/connect/bigquery/schemaregistry/schemaretriever/SchemaRegistrySchemaRetriever.java
```diff
@@ -23,7 +24,9 @@
 
 import java.io.IOException;
 
-import java.util.Map;
+import java.util.*;
```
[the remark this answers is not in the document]
mkubala: Roger that 👍

kcbq-confluent/src/main/java/com/wepay/kafka/connect/bigquery/schemaregistry/schemaretriever/SchemaRegistrySchemaRetriever.java
```diff
@@ -79,9 +103,33 @@ public Schema retrieveSchema(TableId table, String topic, KafkaSchemaRecordType
...
 
   @Override
-  public void setLastSeenSchema(TableId table, String topic, Schema schema) { }
+  public void setLastSeenSchema(TableId table, TopicAndRecordName topicAndRecordName, Schema schema) {
+  }
+
+  private Map<String, TopicAndRecordName> getSubjects(List<String> topics, Map<Pattern, String> recordAliases) {
+    Collection<String> subjects;
+    try {
+      subjects = schemaRegistryClient.getAllSubjects();
```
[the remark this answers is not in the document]
mkubala: Maybe there is a better solution for fetching all supported subjects but unfortunately I could not find it.

We are using Kafka topics to carry on some events, like user logged in, user paid, admin removed an account, etc. Each type of an event has its own record type and we rely on `TopicRecordNameStrategy` (`io.confluent.kafka.serializers.subject.TopicRecordNameStrategy`).
This might sound like a hack while it's one of the fully supported use cases (see https://www.confluent.io/blog/put-several-event-types-kafka-topic/).

Number of types of events (record types) might be humongous and increase in time (usually new feature introduces a couple of events). Maintaining a list of all supported record types for a topic in the connector's configuration would not only be cumbersome but also would require restarts and reconfiguration whenever you introduce new record type (which becomes more problematic when using some hosted kafka solutions).

That's why I decided to rely on regular expressions which let me declare one or a couple of patterns (instead of dozens or hundreds) to rule them all. Also since creating missing tables on the fly is supported, there is no need for restarts.

Unfortunately we cannot query schema reqistry using those patterns and we do not know exact record names in advance.
The only solution is to fetch them all and filter, using regular expressions defined in the config.

Alternatively we could just forget about creating tables at startup time and create them on the fly, when a new record type appears (which I've already implemented in this PR and can rebase on top of the recent changes introduced in #233 ) 🤔 
What do you think?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
mkubala: Yes, there is a reason, described in the previous comment (https://github.com/wepay/kafka-connect-bigquery/pull/219#discussion_r363647317)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConnectorConfig.java
mkubala: The only member of this config (auto create tables) has been moved to its parent (`BigQuerySinkConfig`) because I had to access this property also from the task level.
After that `BigQuerySinkConnectorConfig` became empty.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/convert/BigQueryRecordConverter.java
mkubala: Sometimes particular record types have fields with instances of classes with no fields (null objects, markers etc.).
Record and schema converters should omit such nested empty structs, otherwise BigQuery will raise an error.

mkubala:
Thanks for you feedback!

I've addressed most of the comments. As the next step I'm going to rebase my work on top of @bingqinzhou's work and prepare a few, better grained and smaller PRs.
I'm not going to rebase this PR, since force push will discard all the valuable comments.

I hope that my answers in this PR made my motivation more clear and you will notice the dependencies between all the changes. 
Beside support for handling multiple record types, being able to create tables on the fly and handle nested empty structures in both record and schema converters is essential to make the whole feature reasonably usable on production. 
So you might expect at least 3 smaller PRs from me and it would be much appreciated if they could be released together.

PR #228, opened 2019-12-16
Support ingestion to default partition (#221)

### LINKED ISSUE #221: Support ingestion to default partition
# Background

* In current implementation, BigQuery sink connector specifies partition explicitly like `mydataset.table$20170301` based on timestamp column or current date time. https://github.com/wepay/kafka-connect-bigquery/blob/6270dd8726acab7b897ad4139d22dd2fe90ce836/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java#L144-L146
* When using partition decorator, newly arriving data directly goes to specified partition (but on streaming buffer) instead of `UNPARTITIONED` partition. In this case, we can not exclude newly arriving data when query by filtering out the NULL values from the UNPARTITIONED partition. See the following URL for more details. https://cloud.google.com/bigquery/streaming-data-into-bigquery#streaming_into_ingestion-time_partitioned_tables
* This will cause a problem when we run DML (Data Manipulation Language) due to the following limitation since we can not exclude newly arriving data. https://cloud.google.com/bigquery/docs/reference/standard-sql/data-manipulation-language#limitations
```Rows that were written to a table recently via streaming (using the tabledata.insertall method) cannot be modified using UPDATE, DELETE, or MERGE statements. Recent writes are typically those that occur within the last 30 minutes. Note that all other rows in the table remain modifiable by using UPDATE, DELETE, or MERGE statements.```
* We see the following error when we run a DML while we have data on streaming buffer.
```
Tables that have been written to recently via BigQuery Streaming (tabledata.insertall) cannot be modified using UPDATE or DELETE statements
```

# Proposal
* Add a feature to configure whether (1) or (2). Current implementation is (2)
** (1) do not specify partition
** (2) specify partition
*** (2-1) use timestamp column if useMessageTimeDatePartitioning = true
*** (2-2) use current date time useMessageTimeDatePartitioning = false

### PR DESCRIPTION
**Pull request by yohei1126**
fix #221

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -81,6 +81,7 @@ public class BigQuerySinkTask extends SinkTask {
   private RecordConverter<Map<String, Object>> recordConverter;
   private Map<String, TableId> topicsToBaseTableIds;
   private boolean useMessageTimeDatePartitioning;
+  private boolean explicitPartitioning;
 
   private TopicPartitionManager topicPartitionManager;
 
@@ -138,17 +139,19 @@ public class BigQuerySinkTask extends SinkTask {
     TableId baseTableId = topicsToBaseTableIds.get(record.topic());
 
     PartitionedTableId.Builder builder = new PartitionedTableId.Builder(baseTableId);
-    if (useMessageTimeDatePartitioning) {
-      if (record.timestampType() == TimestampType.NO_TIMESTAMP_TYPE) {
-        throw new ConnectException(
-            "Message has no timestamp type, cannot use message timestamp to partition.");
+
+    if (explicitPartitioning) {
+      if (useMessageTimeDatePartitioning) {
+        if (record.timestampType() == TimestampType.NO_TIMESTAMP_TYPE) {
+          throw new ConnectException(
+                  "Message has no timestamp type, cannot use message timestamp to partition.");
+        }
+
+        builder.setDayPartition(record.timestamp());
+      } else {
+        builder.setDayPartitionForNow();
       }
-
-      builder.setDayPartition(record.timestamp());
-    } else {
-      builder.setDayPartitionForNow();
     }
-
     return builder.build();
   }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by yohei1126** [comment]
@mtagle @criccomini Could you please review.

**Comment by darshanmehta10** [comment]
This is the exact duplicate of https://github.com/wepay/kafka-connect-bigquery/pull/203 (created by me in Nov) and https://github.com/wepay/kafka-connect-bigquery/pull/229 (same as #203 with rebasing the intermediate pushes). Not sure why this was created in the first place.

File: kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java [hunk 1]
```diff
@@ -138,17 +139,19 @@ private PartitionedTableId getRecordTable(SinkRecord record) {
...
-      if (record.timestampType() == TimestampType.NO_TIMESTAMP_TYPE) {
-        throw new ConnectException(
-            "Message has no timestamp type, cannot use message timestamp to partition.");
-      }
 
-      builder.setDayPartition(record.timestamp());
-    } else {
-      builder.setDayPartitionForNow();
-    }
+    if (explicitPartitioning) {
```
darshanmehta10: Message time partitioning is also explicit partitioning (as we are defining the partition with the records). This config name name along with `useMessageTimeDatePartitioning` looks conflicting. Also, it's not following the pattern for the other config parameters (i.e. `useXxx`).

yohei1126: No suggestion for alternative name?
How’s about useExplicitIngestionTimePartitionDecoration?

darshanmehta10: That's too verbose. Another name could be `usePartitionDecorator` but that would mean this would become the exact replica of https://github.com/wepay/kafka-connect-bigquery/pull/229 if it's not the replica already.

mtagle:
I left a comment on the associated issue, but I wanted to mention it here so it's less likely to be missed.

criccomini:
I read the document the same way that @mtagle does. Have you tested this out locally to verify that it works as you expect--that you can exclude rows and run your DML? That would be convincing to me.

PR #237, opened 2020-01-07
Add support for handling empty schemas and records

mkubala:
Sometimes record types contain fields of type that do not have any fields defined (null objects, marker classes etc.).
Converters must omit them since it's impossible to map empty structures into BigQuery table schema and rows.

This is the first of PRs derived from #219


jaroslawZawila:
@mtagle @criccomini @C0urante Any progress on this one?

criccomini:
Will have @mtagle take a look at this shortly

mtagle:
This looks fine, but sort of out of style with the rest of the code base. Can you tell me why using Optional/list comprehensions/other functional stuff is preferable to more "standard" imperative approach?

mkubala:
I've been programming in Scala for last 5-6 years and I get used to optionals and list comprehensions. 

It was just easier to write for me and I thought since Java has the Streams API it's natural descendant of for comprehensions and mutable classes, just like for comprehensions at some point for comps crowded while loops out.
If you prefer to stay with the original approach - I'm ok with it.

mtagle:
hm, in practice I suppose other code will only be interacting with the schema converters in the convertSchema method, which isn't touched, so I think I'm okay with the style variance.

PR #233, opened 2019-12-19
Support non existent topics in KCBQ.

bingqinzhou:
When `autoCreateTables` config is set to true:

- Do not throw any error when the table of a topic doesn't exist in BigQuery upon starting the connector.
- Instead of auto-creating tables upon starting the connector, create tables when we actually receive the kafka sink record.
- The situation is already covered by current integration tests.

PR #238, opened 2020-01-09
Add support for topics with multiple schemas

mkubala:
This contribution makes kafka-connect-bigquery compatible with topics that carry different types of messages, especially when Kafka is used for an event sourcing.

Approach of putting several event types in a single topic is already supported by Kafka and Schema Registry (see https://www.confluent.io/blog/put-several-event-types-kafka-topic/).

This PR has been derived from #219 and rebased on top of the recent changes made by @bingqinzhou

### DIFF
--- a/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
+++ b/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
@@ -4,7 +4,9 @@ import com.google.cloud.bigquery.TableId;
 
 import org.apache.kafka.connect.data.Schema;
 
+import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 /**
  * Interface for retrieving the most up-to-date schemas for a given BigQuery table. Used in
@@ -21,17 +23,19 @@ public interface SchemaRetriever {
   /**
    * Retrieve the most current schema for the given topic.
    * @param table The table that will be created.
-   * @param topic The topic to retrieve a schema for.
+   * @param topicAndRecordName The topic and an optional record name to retrieve a schema for.
    * @param schemaType The type of kafka schema, either "value" or "key".
    * @return The Schema for the given table.
    */
-  public Schema retrieveSchema(TableId table, String topic, KafkaSchemaRecordType schemaType);
+  public Schema retrieveSchema(TableId table, TopicAndRecordName topicAndRecordName, KafkaSchemaRecordType schemaType);
+
+  Map<TopicAndRecordName, Schema> retrieveSchemas(List<String> topics, Map<Pattern, String> recordAliases);
 
   /**
    * Set the last seen schema for a given topic
    * @param table The table that will be created.
-   * @param topic The topic to retrieve a schema for.
+   * @param topicAndRecordName The topic and an optional record name to retrieve a schema for.
    * @param schema The last seen Kafka Connect Schema
    */
-  public void setLastSeenSchema(TableId table, String topic, Schema schema);
+  public void setLastSeenSchema(TableId table, TopicAndRecordName topicAndRecordName, Schema schema);
 }
--- a/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
+++ b/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
@@ -18,7 +18,7 @@ public interface SchemaRetriever {
    * {@link org.apache.kafka.connect.sink.SinkConnector#start(Map)} method.
    * @param properties The configuration settings of the connector.
    */
-  public void configure(Map<String, String> properties);
+  void configure(Map<String, String> properties);
 
   /**
    * Retrieve the most current schema for the given topic.
@@ -27,8 +27,14 @@ public interface SchemaRetriever {
    * @param schemaType The type of kafka schema, either "value" or "key".
    * @return The Schema for the given table.
    */
-  public Schema retrieveSchema(TableId table, TopicAndRecordName topicAndRecordName, KafkaSchemaRecordType schemaType);
+  Schema retrieveSchema(TableId table, TopicAndRecordName topicAndRecordName, KafkaSchemaRecordType schemaType);
 
+  /**
+   * Retrieve the most current schemas that match the configured topics and records.
+   * @param topics The list of a supported topics.
+   * @param recordAliases The map containing record name patterns and their aliases.
+   * @return the most current schemas that match the configured topics and records.
+   */
   Map<TopicAndRecordName, Schema> retrieveSchemas(List<String> topics, Map<Pattern, String> recordAliases);
 
   /**
@@ -37,5 +43,5 @@ public interface SchemaRetriever {
    * @param topicAndRecordName The topic and an optional record name to retrieve a schema for.
    * @param schema The last seen Kafka Connect Schema
    */
-  public void setLastSeenSchema(TableId table, TopicAndRecordName topicAndRecordName, Schema schema);
+  void setLastSeenSchema(TableId table, TopicAndRecordName topicAndRecordName, Schema schema);
 }
--- a/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
+++ b/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
@@ -4,9 +4,7 @@ import com.google.cloud.bigquery.TableId;
 
 import org.apache.kafka.connect.data.Schema;
 
-import java.util.List;
 import java.util.Map;
-import java.util.regex.Pattern;
 
 /**
  * Interface for retrieving the most up-to-date schemas for a given BigQuery table. Used in
@@ -28,14 +26,6 @@ public interface SchemaRetriever {
    * @return The Schema for the given table.
    */
   Schema retrieveSchema(TableId table, TopicAndRecordName topicAndRecordName, KafkaSchemaRecordType schemaType);
-
-  /**
-   * Retrieve the most current schemas that match the configured topics and records.
-   * @param topics The list of a supported topics.
-   * @param recordAliases The map containing record name patterns and their aliases.
-   * @return the most current schemas that match the configured topics and records.
-   */
-  Map<TopicAndRecordName, Schema> retrieveSchemas(List<String> topics, Map<Pattern, String> recordAliases);
 
   /**
    * Set the last seen schema for a given topic
--- a/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
+++ b/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
@@ -64,37 +64,6 @@ public class TopicAndRecordName {
     return Optional.ofNullable(this.recordName);
   }
 
-  /**
-   * Convert topic and record name into the schema registry subject.
-   * Subjects follow topic-recordName format.
-   *
-   * If recordName is not present, "value" will be used.
-   *
-   * @return corresponding schema registry subject.
-   */
-  public String toSubject() {
-    return toSubject(KafkaSchemaRecordType.VALUE);
-  }
-
-  /**
-   * Convert topic and record name into the schema registry subject.
-   * Subjects follow topic-recordName format.
-   *
-   * If recordName is not present, "value" or "key" will be used, depending on the schema type.
-   *
-   * @param schemaType schema type used to resolve full subject when recordName is absent.
-   * @return corresponding schema registry subject.
-   */
-  public String toSubject(KafkaSchemaRecordType schemaType) {
-    String subject = topic;
-    if (recordName == null) {
-      subject += "-" + schemaType.toString();
-    } else {
-      subject += "-" + recordName;
-    }
-    return subject;
-  }
-
   @Override
   public String toString() {
     return topic + "/" + recordName;
--- a/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
+++ b/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
@@ -1,0 +1,70 @@
+package com.wepay.kafka.connect.bigquery.api;
+
+import org.apache.kafka.connect.sink.SinkRecord;
+
+import java.util.Objects;
+import java.util.Optional;
+
+public class TopicAndRecordName {
+  private final String topic;
+  private final String recordName;
+
+  public static TopicAndRecordName from(String topic) {
+    return new TopicAndRecordName(topic, null);
+  }
+
+  public static TopicAndRecordName from(String topic, String recordName) {
+    return new TopicAndRecordName(topic, recordName);
+  }
+
+  public static TopicAndRecordName from(SinkRecord record, boolean includeRecordName) {
+    String maybeRecordName = includeRecordName ? record.valueSchema().name() : null;
+    return from(record.topic(), maybeRecordName);
+  }
+
+  public TopicAndRecordName(String topic, String recordName) {
+    this.topic = topic;
+    this.recordName = recordName;
+  }
+
+  public String getTopic() {
+    return this.topic;
+  }
+
+  public Optional<String> getRecordName() {
+    return Optional.ofNullable(this.recordName);
+  }
+
+  public String toSubject() {
+    return toSubject(KafkaSchemaRecordType.VALUE);
+  }
+
+  public String toSubject(KafkaSchemaRecordType schemaType) {
+    String subject = topic;
+    if (recordName == null) {
+      subject += "-" + schemaType.toString();
+    } else {
+      subject += "-" + recordName;
+    }
+    return subject;
+  }
+
+  @Override
+  public String toString() {
+    return topic + "/" + recordName;
+  }
+
+  @Override
+  public int hashCode() {
+    return this.topic.hashCode();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj instanceof TopicAndRecordName) {
+      TopicAndRecordName that = (TopicAndRecordName) obj;
+      return Objects.equals(getTopic(), that.getTopic()) && Objects.equals(getRecordName(), that.getRecordName());
+    }
+    return false;
+  }
+}
--- a/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
+++ b/kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
@@ -5,23 +5,52 @@ import org.apache.kafka.connect.sink.SinkRecord;
 import java.util.Objects;
 import java.util.Optional;
 
+/**
+ * A container class with separate topic and (optional) record names.
+ */
 public class TopicAndRecordName {
   private final String topic;
   private final String recordName;
 
+  /**
+   * Convenience method for creating a new {@link TopicAndRecordName} with empty record name.
+   *
+   * @param topic the topic.
+   * @return a new {@link TopicAndRecordName} with empty record name.
+   */
   public static TopicAndRecordName from(String topic) {
     return new TopicAndRecordName(topic, null);
   }
 
+  /**
+   * Convenience method for creating a new {@link TopicAndRecordName}.
+   *
+   * @param topic the topic.
+   * @param recordName the record name.
+   * @return a new {@link TopicAndRecordName}.
+   */
   public static TopicAndRecordName from(String topic, String recordName) {
     return new TopicAndRecordName(topic, recordName);
   }
 
+  /**
+   * Convenience method for creating a new {@link TopicAndRecordName} from {@link SinkRecord}.
+   *
+   * @param record the record used to extract topic and (optionally) record name.
+   * @param includeRecordName whether or not to include the record name.
+   * @return a new {@link TopicAndRecordName}.
+   */
   public static TopicAndRecordName from(SinkRecord record, boolean includeRecordName) {
     String maybeRecordName = includeRecordName ? record.valueSchema().name() : null;
     return from(record.topic(), maybeRecordName);
   }
 
+  /**
+   * Create a new {@link TopicAndRecordName}.
+   *
+   * @param topic the topic
+   * @param recordName the record name
+   */
   public TopicAndRecordName(String topic, String recordName) {
     this.topic = topic;
     this.recordName = recordName;
@@ -35,10 +64,27 @@ public class TopicAndRecordName {
     return Optional.ofNullable(this.recordName);
   }
 
+  /**
+   * Convert topic and record name into the schema registry subject.
+   * Subjects follow topic-recordName format.
+   *
+   * If recordName is not present, "value" will be used.
+   *
+   * @return corresponding schema registry subject.
+   */
   public String toSubject() {
     return toSubject(KafkaSchemaRecordType.VALUE);
   }
 
+  /**
+   * Convert topic and record name into the schema registry subject.
+   * Subjects follow topic-recordName format.
+   *
+   * If recordName is not present, "value" or "key" will be used, depending on the schema type.
+   *
+   * @param schemaType schema type used to resolve full subject when recordName is absent.
+   * @return corresponding schema registry subject.
+   */
   public String toSubject(KafkaSchemaRecordType schemaType) {
     String subject = topic;
     if (recordName == null) {
@@ -56,7 +102,7 @@ public class TopicAndRecordName {
 
   @Override
   public int hashCode() {
-    return this.topic.hashCode();
+    return Objects.hashCode(this.topic) + Objects.hashCode(this.recordName);
   }
 
   @Override
--- a/kcbq-confluent/src/main/java/com/wepay/kafka/connect/bigquery/schemaregistry/schemaretriever/SchemaRegistrySchemaRetriever.java
+++ b/kcbq-confluent/src/main/java/com/wepay/kafka/connect/bigquery/schemaregistry/schemaretriever/SchemaRegistrySchemaRetriever.java
@@ -65,7 +65,7 @@ public class SchemaRegistrySchemaRetriever implements SchemaRetriever {
   @Override
   public Schema retrieveSchema(TableId table, TopicAndRecordName topicAndRecordName, KafkaSchemaRecordType schemaType) throws ConnectException {
     String topic = topicAndRecordName.getTopic();
-    String subject = topicAndRecordName.toSubject(schemaType);
+    String subject = getSubject(topicAndRecordName, schemaType);
     logger.debug("Retrieving schema information for topic {} with subject {} and schema type {}", topic, subject, schemaType);
     try {
       SchemaMetadata latestSchemaMetadata = schemaRegistryClient.getLatestSchemaMetadata(subject);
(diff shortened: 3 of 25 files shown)


jaroslawZawila:
@mtagle @criccomini @C0urante Any progress on this one?

mkubala:
@mtagle @criccomini @C0urante  Is there any chance for this feature to be merged and included in the next release?

I do not want to exert pressure on anyone here, but it's been almost 3 months since I started work on this contribution after receiving a positive feedback in #175 and my team is blocked with their work on ML & reporting. 

I'm between a rock and a hard place and if I should start looking for another solution (write my own custom service or connector?) I'd like to know it before the team get really mad.

mtagle:
hey @mkubala, I've asked @whynick1 to take a look at your PR. I'm eager to get this merged as well!

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -109,6 +109,23 @@
       + "capture groups that are referenced in the format string using placeholders (i.e. $1) "
       + "(form of <topic regex>=<format string>)";
 
+  public static final String RECORD_ALIASES_CONFIG =                         "recordAliases";
+  private static final ConfigDef.Type RECORD_ALIASES_TYPE =                  ConfigDef.Type.LIST;
+  private static final ConfigDef.Importance RECORD_ALIASES_IMPORTANCE =      ConfigDef.Importance.MEDIUM;
+  private static final Object RECORD_ALIASES_DEFAULT =                        null;
+  private static final String RECORD_ALIASES_DOC =
+      "A list of mappings from record name regexes to table name postfixes. Note the regex must include "
```
whynick1: Why "postfixes"?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -109,6 +109,23 @@
       + "capture groups that are referenced in the format string using placeholders (i.e. $1) "
       + "(form of <topic regex>=<format string>)";
 
+  public static final String RECORD_ALIASES_CONFIG =                         "recordAliases";
```
whynick1: Can you please explain how we are going to use `recordAliases` to support topics with multiple schemas? 😄

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -448,6 +477,41 @@ protected static void ensureValidMap(String name, List<String> values) {
     return patternList;
   }
 
+  public String getSingleMatch(String value,
```
whynick1: A non-static public getSingleMatch() inside `BigQuerySinkConfig` is a bit confusing. Suggest put it back to `TopicToTableResolver`

kcbq-confluent/src/main/java/com/wepay/kafka/connect/bigquery/schemaregistry/schemaretriever/SchemaRegistrySchemaRetriever.java
```diff
@@ -79,9 +108,33 @@ public Schema retrieveSchema(TableId table, String topic, KafkaSchemaRecordType
   }
 
   @Override
-  public void setLastSeenSchema(TableId table, String topic, Schema schema) { }
+  public void setLastSeenSchema(TableId table, TopicAndRecordName topicAndRecordName, Schema schema) {
+  }
+
+  private Map<String, TopicAndRecordName> getSubjects(List<String> topics, Map<Pattern, String> recordAliases) {
```
whynick1: Can you explain the idea here about getting subjects from `topics` & `recordNamePatterns`?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkConnector.java
```diff
@@ -100,17 +96,14 @@ private SchemaManager getSchemaManager(BigQuery bigQuery) {
...
   }
 
-  private void ensureExistingTables() {
-    BigQuery bigQuery = getBigQuery();
-    Map<String, TableId> topicsToTableIds = TopicToTableResolver.getTopicsToTables(config);
-    for (TableId tableId : topicsToTableIds.values()) {
+  private void ensureExistingTables(
+      BigQuery bigQuery,
+      Map<TopicAndRecordName, TableId> topicsToTableIds) {
+    for (Map.Entry<TopicAndRecordName, TableId> topicToTableId : topicsToTableIds.entrySet()) {
```
whynick1: nit: add extra space

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkConnector.java
```diff
@@ -120,6 +113,19 @@ private void ensureExistingTables() {
     }
   }
 
+  private void ensureExistingTables() {
```
whynick1: nit: should be placed on top of the other one (with params)

kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
```diff
@@ -21,17 +23,19 @@
...
    * @param table The table that will be created.
-   * @param topic The topic to retrieve a schema for.
+   * @param topicAndRecordName The topic and an optional record name to retrieve a schema for.
    * @param schemaType The type of kafka schema, either "value" or "key".
    * @return The Schema for the given table.
    */
-  public Schema retrieveSchema(TableId table, String topic, KafkaSchemaRecordType schemaType);
+  public Schema retrieveSchema(TableId table, TopicAndRecordName topicAndRecordName, KafkaSchemaRecordType schemaType);
+
+  Map<TopicAndRecordName, Schema> retrieveSchemas(List<String> topics, Map<Pattern, String> recordAliases);
```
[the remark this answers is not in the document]
mkubala: This method is aimed to pre-fetch all the current schemas from registry in order to create all the necessary tables. 

If you take a look on the implementation you'll see that I'm fetching all subjects and then do some pattern matching. I do this because number of types of events (record types) might be humongous and increase in time (usually new feature introduces a couple of events). Maintaining a list of all supported record types for a topic in the connector's configuration would not only be cumbersome but also would require restarts and reconfiguration whenever you introduce new record type (which becomes more problematic when using some hosted kafka solutions).

That's why I decided to rely on regular expressions which let me declare one or a couple of patterns (instead of dozens or hundreds) to rule them all. Also since creating missing tables on the fly is supported, there is no need for restarts.

Unfortunately we cannot query schema reqistry using those patterns and we do not know exact record names in advance.
The only solution is to fetch them all and filter, using regular expressions defined in the config.

Alternatively we could just forget about creating tables at startup time and create them on the fly, when a new record type appears. This is something @bingqinzhou introduced recently in #233. If all tables would be created on demand and there won't be need for pre-fetching subjects, then we could get rid of this method and also make several classes (including TopicAndRecordName) much simpler.

WDYT?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -30,51 +35,44 @@
...
   /**
-   * @param schemaRetriever Used to determine the Kafka Connect Schema that should be used for a
-   *                        given table.
-   * @param schemaConverter Used to convert Kafka Connect Schemas into BigQuery format.
    * @param bigQuery Used to communicate create/update requests to BigQuery.
-   * @param kafkaKeyFieldName The name of kafka key field to be used in BigQuery.
-   *                         If set to null, Kafka Key Field will not be included in BigQuery.
-   * @param kafkaDataFieldName The name of kafka data field to be used in BigQuery.
-   *                           If set to null, Kafka Data Field will not be included in BigQuery.
+   * @param config BigQuery sink configuration.
```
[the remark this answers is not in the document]
mkubala: Good point, I'll mention how those are obtained and what happens if they are null in the doc 👍

kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/SchemaRetriever.java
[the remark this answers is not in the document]
whynick1: > Alternatively we could just forget about creating tables at startup time and create them on the fly, when a new record type appears. This is something @bingqinzhou introduced recently in #233. If all tables would be created on demand and there won't be need for pre-fetching subjects, then we could get rid of this method and also make several classes (including TopicAndRecordName) much simpler.

Aha, thanks for the clarification. Creating table on the fly is really what we want to have (and we just added recently). So, yes, I prefer that over creating tables at startup. Sounds like it will also make your implementation easier. (which is great!)

kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return topic + "/" + recordName;
+  }
+
+  @Override
+  public int hashCode() {
+    return this.topic.hashCode();
```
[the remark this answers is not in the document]
mkubala: Good catch!
According to contract between hashCode and equal, two instances with the same hashCode might not be equal. But I forgot that in case of a single topic with many records we will end up with HashMaps that has only one bucket... I'm going to include recordName in hashCode.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
mkubala: It's very probable that I just picked a silly name for this property.

Record names might be very long (think about fully qualified names for classes that are placed deep in the package structure) and we might want to make BigQuery table names short and clear for people who are not core dev team members (like data scientists).

Another purpose for this property was to filter the set of all schemas retrieved from schema registry to only those that are relevant (see `SchemaRetriever.getAllSchemas()`). Since we are going to get rid of `getAllSchemas()` it will be used only to provide a functionality of shortening table names.

All ideas for a better property name are welcome 👍

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -96,6 +94,16 @@ TableInfo constructTableInfo(TableId table, Schema kafkaKeySchema, Schema kafkaV
     return tableInfoBuilder.build();
   }
 
+  public Map<TopicAndRecordName, Schema> discoverSchemas() {
```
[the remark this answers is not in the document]
mkubala: About the `Schema` - you're absolutely right, this was introduced before `String` was replaced with `TopicAndRecordName`. Since the latter provides all the necessary informations, `Schema` is no longer needed.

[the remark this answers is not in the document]
mkubala: BTW - the whole method will be removed as we are going to create tables on the fly.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableResolver.java
```diff
@@ -43,13 +42,19 @@
    * @param config Config that contains properties used to generate the map
    * @return A Map associating Kafka topic names to BigQuery table names.
    */
-  public static Map<String, TableId> getTopicsToTables(BigQuerySinkConfig config) {
+  public static Map<TopicAndRecordName, TableId> getTopicsToTables(BigQuerySinkConfig config) {
+    // Based only on the config we cannot determine what record type is used in which topic.
+    // It will be eventually discovered by handling new messages.
+    if (config.getBoolean(BigQuerySinkConfig.SUPPORT_MULTI_SCHEMA_TOPICS_CONFIG)) {
+      return Maps.newHashMap();
```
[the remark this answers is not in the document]
mkubala: Since schema discovery is going to be removed, I'm keeping this line untouched and changing the code in `BigQuerySinkConnector`.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
mkubala: BigQuery table name will follow the `<topic_table_name>_<record_alias>` template. Maybe I used incorrect wording, so I'm going to change the description and include the aforementioned template to make it clear.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableResolver.java
```diff
@@ -124,36 +141,19 @@ public static void updateTopicToTable(BigQuerySinkConfig config, String topicNam
...
-          match = patternMatcher.replaceAll(formatString);
-          previousPattern = pattern.getKey().toString();
-        } catch (IndexOutOfBoundsException err) {
-          throw new ConfigException("Format string '" + formatString
-              + "' is invalid in property '" + BigQuerySinkConfig.TOPICS_TO_TABLES_CONFIG
-              + "'", err);
-        }
-      }
-    }
+    return config.getSingleMatch(topicName, BigQuerySinkConfig.TOPICS_CONFIG, BigQuerySinkConfig.TOPICS_TO_TABLES_CONFIG);
```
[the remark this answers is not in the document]
mkubala: Since this method depends on the BigQueryConfig only, it seemed to be a kind of a _tangled util method_ - a good candidate to become a part of the config.
Also I thought that somebody might need such functionality in the future and would like to reuse the code.
I can move it back to `TopicToTableResolver` if this argument doesn't convince you.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
mkubala: If arguments listed in https://github.com/wepay/kafka-connect-bigquery/pull/238#discussion_r372871337 won't convince you, I'll do it without any further discussions 👍

kcbq-confluent/src/main/java/com/wepay/kafka/connect/bigquery/schemaregistry/schemaretriever/SchemaRegistrySchemaRetriever.java
mkubala: Before the explanation begins, I'd like to tell that I'm really glad that we are going to create all tables on the fly. Getting rid of this schemas discovery is a relief.

Below is the copy paste from an original PR, answering the same question:
> We are using Kafka topics to carry on some events, like user logged in, user paid, admin removed an account, etc. Each type of an event has its own record type and we rely on TopicRecordNameStrategy (io.confluent.kafka.serializers.subject.TopicRecordNameStrategy).
This might sound like a hack while it's one of the fully supported use cases (see https://www.confluent.io/blog/put-several-event-types-kafka-topic/).
> 
> Number of unique event types (record types) might be humongous and increase in time (usually new feature introduces a couple of events). Maintaining a list of all supported record types for a topic in the connector's configuration would not only be cumbersome but also would require restarts and reconfiguration whenever you introduce new record type (which becomes more problematic when using some hosted kafka solutions).
> 
> That's why I decided to rely on regular expressions which let me declare one or a couple of patterns (instead of dozens or hundreds) to rule them all. Also since creating missing tables on the fly is supported, there is no need for restarts.
> 
> Unfortunately we cannot query schema reqistry using those patterns and we do not know exact record names in advance.
The only solution is to fetch them all and filter, using regular expressions defined in the config.

skyzyx:
# [Codecov](https://codecov.io/gh/wepay/kafka-connect-bigquery/pull/238?src=pr&el=h1) Report
> Merging [#238](https://codecov.io/gh/wepay/kafka-connect-bigquery/pull/238?src=pr&el=desc) into [master](https://codecov.io/gh/wepay/kafka-connect-bigquery/commit/755b86614a087621c694a58628db728948c29df4&el=desc) will **increase** coverage by `0.25%`.
> The diff coverage is `n/a`.


```diff
@@             Coverage Diff              @@
##             master     #238      +/-   ##
============================================
+ Coverage     68.69%   68.95%   +0.25%     
- Complexity      267      273       +6     
============================================
  Files            32       32              
  Lines          1460     1498      +38     
  Branches        152      153       +1     
============================================
+ Hits           1003     1033      +30     
- Misses          409      417       +8     
  Partials         48       48              
```




| [Impacted Files](https://codecov.io/gh/wepay/kafka-connect-bigquery/pull/238?src=pr&el=tree) | Coverage Δ | Complexity Δ | |
|---|---|---|---|
| [...a/connect/bigquery/utils/TopicToTableResolver.java](https://codecov.io/gh/wepay/kafka-connect-bigquery/pull/238/diff?src=pr&el=tree#diff-a2NicS1jb25uZWN0b3Ivc3JjL21haW4vamF2YS9jb20vd2VwYXkva2Fma2EvY29ubmVjdC9iaWdxdWVyeS91dGlscy9Ub3BpY1RvVGFibGVSZXNvbHZlci5qYXZh) | `73.33% <0.00%> (-18.34%)` | `11.00% <0.00%> (-1.00%)` | |

mkubala:
Most of the comments have been addressed. I left some food for thoughts regarding moving the `getSingleMatch` method back from config to the `TopicToTableResolver`.

Also I'd be glad if you could take a look on #237 - it's another feature, essential to start using multi schema topics.  When they finally get merged, I'll expose a third PR, which brings support for resolving datasets based on the topic and record names.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
whynick1: Aha, now I see you why you named it `RECORD_ALIASES_CONFIG`. I think it is fair. 
If you ask for better name, maybe `RECORDS_TO_TABLE_POSTFIXES_CONFIG`?  (similar to `TOPICS_TO_TABLES_CONFIG`) What you think?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/TopicToTableResolver.java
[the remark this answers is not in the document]
whynick1: Make sense to me. Let's keep it here. 😄

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
whynick1: Let's keep this function here. Besides, can you please
1) add doc
2) adjust function param alignment
3) check if it is possible to reuse `getSingleMatch()` in `getSingleMatches()`

kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
```diff
@@ -0,0 +1,116 @@
...
+  /**
+   * Convert topic and record name into the schema registry subject.
+   * Subjects follow topic-recordName format.
+   *
+   * If recordName is not present, "value" or "key" will be used, depending on the schema type.
+   *
+   * @param schemaType schema type used to resolve full subject when recordName is absent.
+   * @return corresponding schema registry subject.
+   */
+  public String toSubject(KafkaSchemaRecordType schemaType) {
```
criccomini: This is leaking Confluent Schema Registry stuff into the API layer, which must be agnostic to specific schema implementations. The concept of a subject is specific to Confluent.

In fact, the entire concept of a record name is specific to the Avro/Confluent implementation. Protobufs will probably work with this since they also have record names, but what about JSON? IIRC, there is another PR that is about supporting JSON schemas.

I am concerned that this PR is far too tied to the Confluent Schema Registry, especially in the API layer.

I am open to changing the API interfaces to make it easier on the Confluent Schema Registry, but not in a way that excludes other potential implementations (especially ones like JSON, where there is active interest). Can you suggest how this can be altered to not leak Confluent Schema Registry assumptions into the API?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
mkubala: It's longer but makes clear what's the purpose of the param 👍

kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
mkubala: Good point!
I'm working now on decoupling this plugin from that particular schema implementation. The rough idea is to extract and encapsulate all the SR-related logic in a specialized classes (so that it should be easy to provide support for different schemas).

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
mkubala: It wasn't possible to reuse getSingleMatch in getSingleMatches because only one of them performs substitution of captured RegEx groups, but I was able to extract the common part of looking up the matching pattern.

kcbq-api/src/main/java/com/wepay/kafka/connect/bigquery/api/TopicAndRecordName.java
mkubala: The responsibility for assembling Confluent Schema Registry subject names has been moved to the SchemaRegistrySchemaRetriever.
Also the `recordName` field stays optional, so in case of an alternative implementations (like aforementioned JSON) won't need to bother about it.

criccomini: Ok, so @whynick1 and I have done some talking and investigating. I think we have a good idea about the best approach to achieve not only what you're trying to do, but to evolve KCBQ to support things in a more generic way in general. This amounts to:

1. Adding a pluggable TableRouter that takes a SinkRecord and returns which table it should be written to. (Default: RegexTableRouter)
2. Changing SchemaRetriever interface to have two methods: `Schema getKeySchema(SinkRecord)` and `Schema getValueSchema(SinkRecord)`. (Default: IdentitySchemaRetriever, which just returns `sinkRecord.keySchema()` and `sinkRecord.valueSchema()`)
3. Changing the way that schema updates are handled in the AdaptiveBigQueryWriter.

This approach should give us a ton of flexibility including:

1. It will allow you to pick which table to route each individual message based on all of the information in SinkRecord (topic, key schema, value schema, message payload, etc.)
2. It will allow us to fix some known schema-evolution bugs in cases where one field is added and another is dropped.
3. It will allow us to use SinkRecord's .keySchema and .valueSchema rather than talking to the schema registry for schemas.
4. It will make it easy to support JSON messages, even those that return null for the .keySchema() and .valueSchema() methods--you can implement a custom retriever for this case.

I am going to write up a GH issue today that documents in detail what needs to be done to implement these changes, and we can go from there.

Does this sound good?

whynick1: Add some note regarding @criccomini 's point 1, 2, 3.
1. As an example to your case,
```
public class MultiSchemaRegexSTableRouter {
  Map<TopicAndRecordName, TableId> topicsToBaseTableIds;

  @override
  public String route(SinkRecord sinkRecord) {
    TopicAndRecordName key = new TopicAndRecordName(sinkRecord.topic(), sinkRecord.recordName());
    if (topicsToBaseTableIds.contains(key) {
      return topicsToBaseTableIds.get(key);
    } else {
      // return matched BQ table with topic & record name
    }
  }
}
```
2. Splitting into `getKeySchema`, `valueSchema` also provide an option to load `SinkRecord`'s key schema (if has) to BQ for various purpose including deduplication, debugging, etc.
3. Automatic schema now is default (should be tunable). If turned on, if a new record uses new schema to insert (to BQ), KCBQ will try to update BQ schema from the latest one from Schema Registry. Now, we want to decouple from that, and instead derive new schema from `Record`, assuming json-like `Record` might not has a schema.

criccomini: See https://github.com/wepay/kafka-connect-bigquery/issues/245

mkubala: I really like this idea!

Yesterday I was struggling with changing an existing code on my local branch, responsible for picking up datasets based on both topic and record name, so that it would keep the API part agnostic and completely unaware of different dataset routing strategies. At the end of the day I was a little bit frustrated, because I had to either add yet another, configurable class (similar to SchemaRetriever but for datasets) and letting plugin users to remember about combining the right schema retriever with the right dataset resolver or add responsibility for routing datasets to SchemaRetriever, thus breaking SRP rule...

The proposed `TableRouter.getTable` method solves all the problems!

criccomini:
@mkubala are you planning to update this with TableRouter?

mkubala:
Hi @criccomini!
Sorry for late response. I cannot tell you if and when I'll find time to update this PR.

I've been working on it as a part of my daily job and since we are behind the original schedule for BigQuery integration I found myself between a rock and a hard place.

My client, who paid for the contribution made so far, does not want to spend more money and see the working integration in our product ASAP.
On the other hand you guys want to keep as clean and maintainable codebase as possible.
Due to personal reasons I cannot sacrifice my spare time in order to introduce `TableRouter` to this PR in a reasonable time. Also I have yet another PR to expose (datasets based on both - topic & record names).

I wonder if there is any chance that you could approve & merge this PR "as it is" and `TableRouter` would be introduced as a separate PR, in a near future?

criccomini:
No, this PR adds too much complexity to the codebase, and isn't the right approach--the refactor is. I don't think it's much more work than this PR, itself. I am reaching out to @rhauch to see if you all can allocate time for refactor that's outlined in issue #245.

mkubala:
I understand. @rhauch please let me know if you start working on that.

rhauch:
@mkubala sorry, I won’t have time to work on this.

mkubala:
There is a light in a tunnel. I spoke to my client this morning and they will let me finish the feature as part of my daily job if I give them a kind of hard estimate / deadline.

How much time you will need to review this PR and release a new version of the plugin? 
Can you guarantee that after this PR got merged (and optionally the other one with support for matching datasets by both topic and record name) a new version of the connector plugin will be released?

mustosm:
Hello,
We are very interested in this feature. Can you tell us when you would consider adding @mkubala's contribution to your release?

mkubala:
@mustosm Maybe I'll find some spare time in the next couple of weeks to finish this feature.
Although I have no idea how much time it will take to re-review this PR and release a new version, so if you cannot wait then build connector from this branch on your own - that's what we did.

mustosm:
@mkubala thank you for your answer. If the feature is released by wepay it would be better. 
Are you using this feature in production today?

mkubala:
It's been used on the staging environment.

OuesFa:
Hi @mkubala 👋 
Any update on this? 
Are you using it in prod on your side please?

OuesFa:
Hi @mkubala 👋 
Any update on this please? 
Are you using this in prod env?
Did you make a release somewhere?

OuesFa:
👋 
I'm sharing a solution for this using SMTs with the current version of the connector. 
https://github.com/confluentinc/kafka-connect-bigquery/issues/114#issuecomment-973118252

PR #242, opened 2020-01-23
Update keyfile config to be a PASSWORD

ilanjiR:
The keyfile config right now is a string. This means that the connector logs are printing the keyfile which is unsafe.

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
@@ -133,7 +133,7 @@ public class BigQuerySinkConfig extends AbstractConfig {
       "A class that can be used for automatically creating tables and/or updating schemas";
 
   public static final String KEYFILE_CONFIG =                     "keyfile";
-  private static final ConfigDef.Type KEYFILE_TYPE =              ConfigDef.Type.STRING;
+  private static final ConfigDef.Type KEYFILE_TYPE =              ConfigDef.Type.PASSWORD;
   public static final String KEYFILE_DEFAULT =                    null;
   private static final ConfigDef.Importance KEYFILE_IMPORTANCE =  ConfigDef.Importance.MEDIUM;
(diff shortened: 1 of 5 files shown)


MosheBlumbergX:
@mtagle , @wicknicks any chance you can check this out please?

wicknicks:
@MosheBlumbergX we need to fix com.wepay.kafka.connect.bigquery.SinkPropertiesFactory#testProperties that tries to read keyfile as a String (instead of the Password). also, it would be good to add some check to ensure that the config() method in BigQuerySinkConnector behaves correctly. Currently, it still reads keyfile as if it were a string, and there doesn't seem to a test catching that problem.

criccomini:
Agree with @wicknicks --this PR looks useful, but needs to pass tests and fix issues mentioned above

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -406,6 +406,13 @@ protected static void ensureValidMap(String name, List<String> values) {
     }
   }
 
+  /**
+   * Returns the keyfile
+   */
+  public String getKeyFile() {
+    return getPassword(KEYFILE_CONFIG).value();
```
C0urante: Will this NPE if no keyfile is specified?

ilanjiR: No, it calls 
```public Password getPassword(String key) {
        return (Password) get(key);
    }
``` 
which calls 
```
protected Object get(String key) {
        if (!values.containsKey(key))
            throw new ConfigException(String.format("Unknown configuration '%s'", key));
        used.add(key);
        return values.get(key);
    }
``` 
so we should be safe if the config is empty

ilanjiR: We throw an exception upstream if the config is null/empty, so I think that should prevent the NPE?

bingqinzhou:
Looks good to me, will merge it.

PR #262, opened 2020-04-15
Fail the task if error is encountered when writing to BigQuery 

Cyril-Engels:
Background: 
The connector writes to BigQuery using a separate thread pool, which means that errors encountered during those writes don’t immediately cause the task to fail. Those errors aren’t swallowed silently, however, and do cause the task to throw an exception from its [`flush method`](https://github.com/wepay/kafka-connect-bigquery/blob/4fb38a8f4b3b6db065976639c57106db9896f0db/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java#L127) when invoking [`KCBQThreadPoolExecutor::awaitCurrentTasks`](https://github.com/wepay/kafka-connect-bigquery/blob/4fb38a8f4b3b6db065976639c57106db9896f0db/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/KCBQThreadPoolExecutor.java#L76-L100).

Unfortunately, the framework does not fail tasks when they throw exceptions from their [`flush method`](https://github.com/apache/kafka/blob/73ec7304b9e13cf4b9da05f215bd356a84efe0e7/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L384-L397) (which is indirectly called as a result of invoking SinkTask::preCommit), and instead, [logs the error and resets the consumer to the last successful offset](https://github.com/apache/kafka/blob/73ec7304b9e13cf4b9da05f215bd356a84efe0e7/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/WorkerSinkTask.java#L384-L397).

This commit exposes the Exception in `BigQuerySinkTask::put`, so the task can fail without keeping reading the same batch of problematic records.


Cyril-Engels:
@mtagle @criccomini Please let me know what you folks think.

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTaskTest.java
```diff
@@ -297,11 +297,43 @@ public void testPutWhenPartitioningOnMessageTimeWhenNoTimestampType() {
...
+
+    SchemaRetriever schemaRetriever = mock(SchemaRetriever.class);
+    SchemaManager schemaManager = mock(SchemaManager.class);
+
+    BigQuerySinkTask testTask = new BigQuerySinkTask(bigQuery, schemaRetriever, storage, schemaManager);
+    testTask.initialize(sinkTaskContext);
+    testTask.start(properties);
+
+    testTask.put(Collections.singletonList(spoofSinkRecord(topic)));
+    while (true) {
```
mtagle, line 326: I appreciate that this test function has a timeout, but can't we just have this run twice rather than sort of arbitrarily large number of times?

kcbq-connector/src/test/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTaskTest.java
```diff
@@ -297,11 +297,43 @@ public void testPutWhenPartitioningOnMessageTimeWhenNoTimestampType() {
         TimestampType.NO_TIMESTAMP_TYPE, null)));
   }
 
-  // It's important that the buffer be completely wiped after a call to flush, since any execption
-  // thrown during flush causes Kafka Connect to not commit the offsets for any records sent to the
-  // task since the last flush
-  @Test
-  public void testBufferClearOnFlushError() {
+  @Test(expected = BigQueryConnectException.class, timeout = 60000L)
```
mtagle, line 300: could you add a comment explaining this test?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -184,6 +184,11 @@ private String getRowId(SinkRecord record) {
 
   @Override
   public void put(Collection<SinkRecord> records) {
+    // check for non-retriable errors and fail the task if any.
+    // adding this because any Exception thrown in flush will be ignored by the framework, which
```
mtagle, line 188: make the comment more relevant to when it'll eventually be committed by explaining why we want to throw an exception in put rather than why it's not sufficient to throw an exception in flush.

PR #263, opened 2020-04-16
Suppress logs for number of records read.

apoorvmittal10:
- The logs generated by `put` method are enormous and not much useful. Moved the logs for `put` records to `debug`.


apoorvmittal10:
@mtagle @wicknicks Need +1 for merge.

criccomini:
@apoorvmittal10 talked it over with our SREs. Prefer keeping this at info. It's useful after the fact. We can't retroactively turn on debug when something bad has already happened. Having it at info is useful in such cases.

If it's really annoying, users can filter it out using log4j filters (or whatever slf4j plugin they're using).

PR #270, opened 2020-05-18
Only list gcs folder blobs in GCSToBQLoadRunnable

zachary-povey:
Functionality for specifying a GCS folder within the bucket to dump objects into was added [here](https://github.com/wepay/kafka-connect-bigquery/pull/164). 

However we ran into a problem when using the GCS->BQ batch mode of the connector; each connector schedules it's own instance of the `GCSToBQLoadRunnable` which does not use the folder when listing objects to load into BigQuery. Because of this, if you have multiple connectors using the same bucket but different folders they all load all the data and so you receive many duplicates and have trouble when deleting gcs objects.

I have passed through the folder prefix as a list option to the `GCSToBQLoadRunnable`, and all tests seem to be passing for me with the update. However a new integration test should added to recreate the issue and confirms this works fixes it. However this is not a straight forward addition of data and schema, as detailed in the readme section for adding integration tests. I thought it was best raised here for how it should be implemented however I would imagine it working as follows:
- Add some objects in same GCS bucket but in a folder next to the target folder of connector
- Run connector
- Assert rows in target folder appeared
- Assert rows in other folder did not appear

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
@@ -350,6 +350,7 @@ public class BigQuerySinkTask extends SinkTask {
     logger.info("Attempting to start GCS Load Executor.");
     gcsLoadExecutor = Executors.newScheduledThreadPool(1);
     String bucketName = config.getString(config.GCS_BUCKET_NAME_CONFIG);
+    String folderName = config.getString(config.GCS_FOLDER_NAME_CONFIG);
     Storage gcs = getGcs();
     // get the bucket, or create it if it does not exist.
     Bucket bucket = gcs.get(bucketName);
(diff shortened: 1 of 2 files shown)


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -359,7 +360,8 @@ private void startGCSToBQLoadTask() {
       BucketInfo bucketInfo = BucketInfo.of(bucketName);
       bucket = gcs.create(bucketInfo);
     }
-    GCSToBQLoadRunnable loadRunnable = new GCSToBQLoadRunnable(getBigQuery(), bucket);
+    Storage.BlobListOption folderPrefixOption = Storage.BlobListOption.prefix(folderName);
+    GCSToBQLoadRunnable loadRunnable = new GCSToBQLoadRunnable(getBigQuery(), bucket, folderPrefixOption);
```
mtagle, line 364: does this work as expected if there is no folder prefix configured?

zachary-povey: I believe so. As the `GCS_FOLDER_NAME_DEFAULT` on the config is set to an empty string, we are filtering for objects with a path prefixed by an empty string which is the same as no filter at all. 

At least that was how I thought it would work, and took the passing integration tests as confirmation :)

criccomini:
@zachary-povey your proposed integration test flow sounds good

zachary-povey:
Hi guys - i've added the integration test as described- it felt a bit hacky (I don't do much java) so feel free to make suggestions for refactoring.

On a different note are you aware of the integration test ever having race conditions? I was having intermittent issues where tables were the tables populated by streaming inserts wouldn't get created but on a re-run the problem would go away...

PR #284, opened 2020-06-25
Throw original error if batch size is reduced to be smaller than 1

cyrusv:
## Problem
When we get responses from BQ that seem like batch size too large issues (but this is not very accurate logic), we cut the batch size, until we get to zero, then throw ConnectException:
```
    if (currentBatchSize == 1) {
      // todo correct exception type?
      throw new ConnectException("Attempted to reduce batch size below 1.");
    }
```
In the case where we mis-diagnose the error response from BQ, this logic obscures the actual response. In practice, I have production issues that are not batch issues, that show up in my logs and exceptions as batch issues because of this logic.

## Solution
The quick-fix here is to throw the original BQ exception instead of trying to reduce batch size below 1, to at least make reading the logs and debugging easier. The deeper fix would be to actually improve the `isBatchSizeError` function, but I don't think we'll always plug any hole, so this PR is a definite improvement, and improvements to `isBatchSizeError` method can be improved in future PRs.

## Refs
https://cloud.google.com/bigquery/docs/error-messages
https://cloud.google.com/bigquery/quotas

### DIFF
--- a/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
+++ b/kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
@@ -84,6 +84,9 @@ public class TableWriter implements Runnable {
           logger.warn("Could not write batch of size {} to BigQuery.", currentBatch.size(), err);
           if (isBatchSizeError(err)) {
             failureCount++;
+            if (currentBatchSize == 1) {
+              throw err;
+            }
             currentBatchSize = getNewBatchSize(currentBatchSize);
           }
         }


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
```diff
@@ -84,6 +84,9 @@ public void run() {
           logger.warn("Could not write batch of size {} to BigQuery.", currentBatch.size(), err);
           if (isBatchSizeError(err)) {
```
C0urante, line 85: It'd be nice if we could update this method to be more accurate as well to avoid unnecessarily chopping the batch size in half every time an unknown error occurs. Anybody know why the `isBatchSizeError` method is so broad and susceptible to false positives?

cyrusv: I think that's a good improvement, but can we take it as a separate issue? The issue solved here is just to give better feedback in the exception trace, so when you ask for task status with REST API you get something actionable.

cyrusv:
@bingqinzhou, @C0urante , @mtagle -- Hoping to get some feedback on this one, I think it is pretty straight-forward improvement in error handling

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
C0urante: I'm mostly wondering because if this exception looks like a batch size error to the the connector, I'm skeptical of how much value we'd get out of throwing it anyways. And if we know what it'd look like and that it would be valuable to see that exception in a stack trace, we should put that knowledge to good use here if we can and clean up another similarly-affected part of the code base.

cyrusv: I see the confusion now -- the point here is to show the error as it was returned from BQ SDK, when `isBatchSizeErrror(...)` returns true for errors that are mis-diagnosed as batch size errors, since this block in `getNewBatchSize` is not that useful:
```
    if (currentBatchSize == 1) {
      // todo correct exception type?
      throw new ConnectException("Attempted to reduce batch size below 1.");
    }
```

C0urante: Right, my thinking is that if the error as it's returned from the BigQuery SDK is really going to be more useful than the `ConnectException` that's thrown right now, it's probably going to be because it's not actually related to the batch size of the request. And if that's the case, since we expect the error to be different, can we use these expectations to patch the `isBatchSizeError` method and give it finer granularity?

cyrusv: No, because Google doesn't give us enough information to say with 100% confidence it is a batch size error. I prefer to keep this PR a tight improvement. If you want to fix the isBatchSizeError, let's do it as a separate PR.

Refs
https://cloud.google.com/bigquery/docs/error-messages
https://cloud.google.com/bigquery/quotas

C0urante: It's not about saying with 100% confidence that something is a batch size error, it's about leveraging our expectations for what errors will be shown to us with this improvement to identify cases where we can confidently say something _isn't_ a batch size error. If you expect the errors exposed with this PR to help identify, say, schema mismatch exceptions, it makes sense to proactively use that knowledge here to improve the `isBatchSizeError` method to avoid continuing to make unnecessary requests to BigQuery that we know will fail, and help surface errors with the connector faster (especially in the case that someone has configured it with very liberal retry settings).

cyrusv: This sounds like a nice thing to have, but let's keep it a separate PR. I don't want to introduce potential complication to what should be a very simple improvement.

C0urante: Why save it for a separate PR? If there are concrete issues you've been experiencing that would make this improvement worthwhile, that should give you at least a starting point to work off of for improving the `isBatchSizeError` method. It doesn't have to be perfect but any improvement here would be beneficial if you already know what kinds of errors are possibly going to be encountered (which it sounds like you do).

Issue #285, opened 2020-06-29
Replace Docker-based integration tests with embedded integration tests


C0urante:
When we wrote the Docker-based integration tests, there were no options for running an embedded instance of the Connect framework in a unit test. Now, there's the [EmbeddedConnectCluster](https://github.com/apache/kafka/blob/2dfb6b05e1fd697b085f937ddfb70ea28854b8ff/connect/runtime/src/test/java/org/apache/kafka/connect/util/clusters/EmbeddedConnectCluster.java) class provided and supported by Kafka that we can leverage for integration tests, and Confluent also provides [an option](https://github.com/confluentinc/schema-registry/blob/d4e569eea56a4d403c6e41d8a7b99dcbccf85bad/core/src/test/java/io/confluent/kafka/schemaregistry/RestApp.java) for bringing up an embedded instance of Schema Registry.

Using both of these, it should be possible to recreate the existing tests very closely but with an improved workflow for running and expanding them that doesn't require shell scripts or Docker image building, and also enables more flexibility in the test cases covered. We might build on this to add specific tests for upsert/delete if/when that's added to the connector.

Issue #264, opened 2020-04-16
Add upsert/delete support via periodic MERGE statements


C0urante:
Although it's not a common use case for BigQuery, it's come up that it'd be nice to support upserts and deletes with this connector. I'd like to propose the following way to allow the connector to optionally support that functionality, without negatively impacting existing users of the connector in append-only mode:

1. Derive the key fields for the upsert and/or delete from the key of the message in Kafka, which will be required to be either a map or a struct. This will allow us to use the field names from that key and map them to columns in the destination BigQuery table.
2. Insert all records received by the connector into an intermediate table, possibly named something like `${table}_tmp`, where `${table}` is the name of the table that the records are destined for.
3. Periodically perform a `MERGE` operation that uses the intermediate table as its source and includes logic for upsert (if the value of the record is non-null) and/or delete (if the value of the record is null) based on the columns present in the record key.

This should only take place if the user enables upserts or deletes; otherwise, the connector will continue streaming directly to the intended table (or GCS bucket) as it does now.

There's definitely some details that need to be ironed out here, such as what will happen if the fields in the record key change over time, how this would interact with table partitioning, and how/where the `MERGE` would be performed, but this is the overall idea.

@mtagle, @criccomini, what are your thoughts on this?

C0urante:
@mtagle @criccomini bump :)
Would you guys be interested in this and willing to review a PR for it?

criccomini:
🧠 (thinking)

whynick1:
If this feature is to prevent duplicate records, it is not very helpful, because we can add a view on top of BQ table with deduplication logic. 

If this feature is to help with table compaction. Then I see value of it. (if too many duplicates exists in BQ, query table without compaction might be costly in both time and money)

But I also have some concerns about [MERGE](https://cloud.google.com/bigquery/docs/reference/standard-sql/dml-syntax#merge_statement). As one of the UML statement Google provides, there is a quota limit of 1500 to 4000 per day.

Even we take full advantage of it, still means KCBQ connector (under upsert/delete mode) will no longer be real-time streaming. Meanwhile, I suspect if want to do so, because BQ(OLAP) is deigned for append-only. Though UML statements available, it's kind of mis-use to run them frequently.

Last but not least, as intermediate table grow, MERGE become more expensive. Are we going to clear it at any point, like after each MERGE complete? If so, would that affect data streaming?

criccomini:
[...]

PR #286, opened 2020-06-29
GH-264: Add support for upsert/delete via periodic merge flushes

C0urante:
Addresses #264

Implements upsert/delete functionality as proposed in the [collaborative design doc](https://docs.google.com/document/d/1p8_rLQqR9GIALIruB3-MjqR8EgYdaEw2rlFF1fxRJf0/edit#), fixes some bugs in the connector, and adds an embedded integration test framework and tests for upsert/delete.

Some notes:

- The README is updated to include instructions on running the new embedded integration tests. It should be fairly straightforward to run these if you're used to running the existing Docker-based integration tests as the same environment variables are used and, beyond that, all that's required is a Gradle one-liner.
- The integration testing framework added here can be expanded to include new tests and possibly address #285
- The `SchemaRetriever` API has been updated to differentiate between key and value schemas when invoking `setLastSeen`; this is done in a backwards-compatible fashion, so existing implementations of that interface should work with the updated API seamlessly.
- The `MemorySchemaRetriever` is updated to use that new API correctly (the existing one didn't differentiate between key and value schemas, which led to some buggy behavior that was unearthed during integration testing)
- The `SchemaManager` class is optimized to avoid redundant operations (which is useful in general but also specifically with upsert/delete enabled as create/update requests are much more likely in that case and the likelihood of them being redundant is fairly high)
- `BigQuerySinkTask` is tweaked to avoid creating multiple `BigQuery` and `SchemaManager` instances; both classes are thread-safe, so this should be fine
- Existing unit tests are modified to reflect some of these tweaks
- New unit tests are added for other changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,12 +131,41 @@ adjusting flags given to the Avro Console Producer and tweaking the config setti
 
 ## Integration Testing the Connector
 
+There is a legacy Docker-based integration test for the connector, and newer integration tests that
+programmatically instantiate an embedded Connect cluster.
+
+### Embedded integration tests
+
+Currently these tests only verify the connector's upsert/delete feature. They should eventually
+replace all of the existing Docker-based tests.
+
+#### Configuring the tests
+
+You must supply the following environment variables in order to run the tests:
+
+- `$KCBQ_TEST_PROJECT`: The name of the BigQuery project to use for the test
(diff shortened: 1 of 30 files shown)


C0urante:
cc @whynick1 @mtagle @criccomini 

This is a big one. No rush, but it's up now if you want to take a look.

Issue #245, opened 2020-02-04
Decouple topics from tables and generalize SchemaRetriever


criccomini:
# Intro

We have had a number of PRs and issues recently that are attempting to do two things:

1. Support more serialization formats
2. Support multiple schemas in the same topic

These include:

https://github.com/wepay/kafka-connect-bigquery/pull/238
https://github.com/wepay/kafka-connect-bigquery/issues/216
https://github.com/wepay/kafka-connect-bigquery/issues/175
https://github.com/wepay/kafka-connect-bigquery/issues/206
https://github.com/wepay/kafka-connect-bigquery/issues/178

And more.

# Changes

I believe that we can address these issues with the following changes:

### Add TableRouter
1. Adding a pluggable TableRouter that takes a SinkRecord and returns which table it should be written to. (Default: RegexTableRouter)

```
public interface TableRouter {
  void configure(Map<String, String> properties);
  TableId getTable(SinkRecord sinkRecord);
}
```

### Generalize SchemaRetriever
2. Changing SchemaRetriever interface to have two methods: Schema getKeySchema(SinkRecord) and Schema getValueSchema(SinkRecord). (Default: IdentitySchemaRetriever, which just returns sinkRecord.keySchema() and sinkRecord.valueSchema())

```
public interface SchemaRetriever {
  void configure(Map<String, String> properties);
  Schema retrieveKeySchema(SinkRecord sinkRecord);
  Schema retrieveValueSchema(SinkRecord sinkRecord);
}
```

### Change *QueryWriter schema update logic
3. Changing the way that schema updates are handled in the AdaptiveBigQueryWriter.

This change will be the largest. If we remove `SchemaRetriever.updateSchema`, we need a way for AdaptiveBigQueryWriter to update BQ schemas when a batch insert fails. Given these rules:

https://cloud.google.com/bigquery/docs/managing-table-schemas

>>>
This document describes how to modify the schema definitions for existing BigQuery tables. BigQuery natively supports the following schema modifications:

* Adding columns to a schema definition
* Relaxing a column's mode from REQUIRED to NULLABLE
* It is valid to create a table without defining an initial schema and to add a schema definition to the table at a later time.

All other schema modifications are unsupported and require manual workarounds, including:

* Changing a column's name
* Changing a column's data type
* Changing a column's mode (aside from relaxing REQUIRED columns to NULLABLE)
* Deleting a column
>>>

[...]

PR #291, opened 2020-07-28
Generalize schema retriever

sahithi03:
Changes made according to the Github issue - #245 
* Added regexRouter SMT which replaces the topic field of each SinkRecord with (dataset)=(tablename)
* Removed the topicToTableResolver class since it will no longer be used. Refactored and adjusted behavior accordingly to make use of sinkRecord with topic field containing destination table.
*  SchemaRetriever API changed to retrieve the schema from each SinkRecord.
*  Removed SchemaRegistrySchemaRetriever and MemorySchemaRetriever as these retrieve schema based on the topic and wouldn't be compatible with the new API and SMTs.
*  Implemented Schema Unionization logic to retrieve the schema from a batch of sink records with heterogenous schemas and for Schema Updates - Currently handles only two cases:
    * Adding new fields to the schema
    * Changing a field's mode from Required to Nullable.
* Updated Java doc
* Integration tests needed some changes to work with SMTs and schema unionization.
* Unit tests yet to be fixed.


kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -133,15 +132,27 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
...
+      throw new BigQueryConnectException("Table '" + table + "' does not exist");
     }
+  }
 
-    TableId baseTableId = topicsToBaseTableIds.get(record.topic());
+  private PartitionedTableId getRecordTable(SinkRecord record) {
+
+    // SMT RegexTransformation replaces the topic with <Dataset>=<TableName>
+    String dataset = record.topic().split("=")[0];
+    String tableName = record.topic().split("=")[1];
```
C0urante, line 148: Is there heavy demand for dynamic routing to different datasets? An alternative here could be to allow the user to configure a single static dataset in the connector and, if multiple datasets are desired, use multiple connector instances.

Or, for something in the middle, we can allow the user to specify a static dataset as a "default" in case the record topic doesn't use the `<dataset>=<table>` syntax.

What do you think?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -133,15 +132,27 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
     topicPartitionManager.resumeAll();
   }
 
-  private PartitionedTableId getRecordTable(SinkRecord record) {
-    // Dynamically update topicToBaseTableIds mapping. topicToBaseTableIds was used to be
-    // constructed when connector starts hence new topic configuration needed connector to restart.
-    // Dynamic update shall not require connector restart and shall compute table id in runtime.
-    if (!topicsToBaseTableIds.containsKey(record.topic())) {
-      TopicToTableResolver.updateTopicToTable(config, record.topic(), topicsToBaseTableIds);
+  private void ensureExistingTable(TableId table) {
```
C0urante, line 135: We could rename this to `maybeEnsureExistingTable` and then do the check for `config.getBoolean(config.TABLE_CREATE_CONFIG)` here instead of putting that burden on the caller.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -356,8 +336,11 @@ private void startGCSToBQLoadTask() {
     if (bucket == null) {
       // todo here is where we /could/ set a retention policy for the bucket,
       // but for now I don't think we want to do that.
-      BucketInfo bucketInfo = BucketInfo.of(bucketName);
-      bucket = gcs.create(bucketInfo);
+      if (config.getBoolean(config.AUTO_CREATE_BUCKET_CONFIG)) {
+        BucketInfo bucketInfo = BucketInfo.of(bucketName);
+        bucket = gcs.create(bucketInfo);
+      }
+      else throw new ConfigException("Bucket does not exist. Set "+ config.AUTO_CREATE_BUCKET_CONFIG + " to true");
```
C0urante, line 343: Suggestion:
```suggestion
      else throw new ConnectException("Bucket '" + bucketName + "' does not exist; create the bucket manually, or set '" + config.AUTO_CREATE_BUCKET_CONFIG + "' to true.");
```

kcbq-connector/quickstart/properties/connector.properties
```diff
@@ -22,8 +22,7 @@ sanitizeTopics=true
 
 autoUpdateSchemas=true
 
-schemaRetriever=com.wepay.kafka.connect.bigquery.schemaregistry.schemaretriever.SchemaRegistrySchemaRetriever
-schemaRegistryLocation=http://localhost:8081
+schemaRetriever=com.wepay.kafka.connect.bigquery.retrieve.IdentitySchemaRetriever
```
C0urante, line 26: Should we add the `RegexRouter` SMT here so users have an example of how to use it?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -67,32 +68,90 @@ public SchemaManager(
...
-    Schema kafkaValueSchema = schemaRetriever.retrieveSchema(table, topic, KafkaSchemaRecordType.VALUE);
-    Schema kafkaKeySchema = kafkaKeyFieldName.isPresent() ? schemaRetriever.retrieveSchema(table, topic, KafkaSchemaRecordType.KEY) : null;
-    bigQuery.create(constructTableInfo(table, kafkaKeySchema, kafkaValueSchema));
+  public void createTable(TableId table, Set<SinkRecord> records) {
+    List<com.google.cloud.bigquery.Schema> bigQuerySchemas = new ArrayList<>();
+    String tableDescription = null;
+    for (SinkRecord record: records) {
+      Schema kafkaValueSchema = schemaRetriever.retrieveValueSchema(record);
+      Schema kafkaKeySchema = kafkaKeyFieldName.isPresent() ? schemaRetriever.retrieveKeySchema(record) : null;
+      tableDescription = (kafkaValueSchema.doc() != null) ? kafkaValueSchema.doc() : tableDescription;
```
C0urante, line 79: Nit: don't need the parentheses here
```suggestion
      tableDescription = kafkaValueSchema.doc() != null ? kafkaValueSchema.doc() : tableDescription;
```

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -233,6 +223,28 @@
   private static final String TABLE_CREATE_DOC =
           "Automatically create BigQuery tables if they don't already exist";
 
+  public static final String AUTO_CREATE_BUCKET_CONFIG =                "autoCreateBucket";
+  private static final ConfigDef.Type AUTO_CREATE_BUCKET_TYPE =          ConfigDef.Type.BOOLEAN;
+  public static final Boolean AUTO_CREATE_BUCKET_DEFAULT =                true;
+  private static final ConfigDef.Importance AUTO_CREATE_BUCKET_IMPORTANCE = ConfigDef.Importance.MEDIUM;
+  private static final String AUTO_CREATE_BUCKET_DOC =
+          "Whether to automatically create the given bucket, if it does not exist";
```
C0urante, line 231: This only applies when GCS batch loading is enabled, right? If so, can we briefly clarify that here?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -233,6 +223,28 @@
...
+
+  public static final String ADD_NEW_BQ_FIELDS_CONFIG =                 "addNewBigQueryFields";
+  private static final ConfigDef.Type ADD_NEW_BQ_FIELDS_TYPE =          ConfigDef.Type.BOOLEAN;
+  public static final Boolean ADD_NEW_BQ_FIELDS_DEFAULT =               false;
+  private static final ConfigDef.Importance ADD_NEW_BQ_FIELDS_IMPORTANCE = ConfigDef.Importance.MEDIUM;
+  private static final String ADD_NEW_BQ_FIELDS_DOC =
+          "If true, new fields can be added to BigQuery tables during subsequent schema updates"
```
C0urante, line 238: How is this different from the existing schema update behavior?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/retrieve/IdentitySchemaRetriever.java
```diff
@@ -0,0 +1,28 @@
+package com.wepay.kafka.connect.bigquery.retrieve;
```
C0urante, line 1: Don't forget the copyright header :)

kcbq-connector/test/resources/connector-template.properties
```diff
@@ -26,9 +26,12 @@ maxWriteSize=1000
...
-schemaRegistryLocation=http://schema-registry:8081
+schemaRetriever=com.wepay.kafka.connect.bigquery.retrieve.IdentitySchemaRetriever
 
 enableBatchLoad=kcbq_test_gcs-load
 batchLoadIntervalSec=10
 
+transforms=RegexTransformation
+transforms.RegexTransformation.type=org.apache.kafka.connect.transforms.RegexRouter
+transforms.RegexTransformation.regex=.*
+transforms.RegexTransformation.replacement=$0
```
C0urante, line 37: What's this supposed to do? AFAICT it's going to grab the entire topic (matched by `.*`), and then replace it with the entire match (`$0`)... so will this have any effect on the records here?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
stoynov96: This sounds good to me

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
```diff
@@ -233,6 +223,28 @@
...
+          + "and would be NULLABLE";
+
+  public static final String CHANGE_REQ_BQ_FIELDS_TO_NULLABLE_CONFIG =   "changeRequiredFieldsToNullable";
```
criccomini, line 241: This is a confusing name. We have a separate config (`allBQFieldsNullable`) that forces all required fields to nullable. Something like `allowRequiredFieldRelaxation` or something.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
criccomini: Bike shedding, but I think `allowNewBigQueryFields` is more explicit.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
criccomini: Related: What if `allBQFieldsNullable=true`, and `allowRequiredFieldRelaxation=false`? This seems like a irresolvable conflict--the connector would be configured to set all fields to nullable, but also is configured not to allow that.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/retrieve/IdentitySchemaRetriever.java
```diff
@@ -0,0 +1,28 @@
...
+import com.wepay.kafka.connect.bigquery.api.SchemaRetriever;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.sink.SinkRecord;
+
+import java.util.Map;
+
+/**
+ * Fetches the key Schema and value Schema from a Sink Record
+ */
+
```
criccomini, line 12: Nit: remove newline

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/retrieve/IdentitySchemaRetriever.java
criccomini: (On all new files)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/SinkRecordConverter.java
```diff
@@ -0,0 +1,47 @@
...
+import com.wepay.kafka.connect.bigquery.convert.RecordConverter;
+import org.apache.kafka.connect.sink.SinkRecord;
+
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * A class for converting a {@link SinkRecord SinkRecord} to {@link InsertAllRequest.RowToInsert BigQuery row}
+ */
+
```
criccomini, line 16: Nit: remove newline

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/SinkRecordConverter.java
```diff
@@ -0,0 +1,47 @@
...
+public class SinkRecordConverter {
+    private BigQuerySinkTaskConfig config;
+    private RecordConverter<Map<String, Object>> recordConverter;
+
+    public SinkRecordConverter(BigQuerySinkTaskConfig config) {
+        this.config = config;
+        this.recordConverter = config.getRecordConverter();
+    }
+    public InsertAllRequest.RowToInsert getRecordRow(SinkRecord record) {
```
criccomini, line 25: Nit: newline

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/SinkRecordConverter.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * A class for converting a {@link SinkRecord SinkRecord} to {@link InsertAllRequest.RowToInsert BigQuery row}
+ */
+
+public class SinkRecordConverter {
+    private BigQuerySinkTaskConfig config;
+    private RecordConverter<Map<String, Object>> recordConverter;
+
+    public SinkRecordConverter(BigQuerySinkTaskConfig config) {
+        this.config = config;
```
criccomini, line 22: Would prefer this class takes recordConverter, boolean sanitizeFieldName, (optional) keyName, (optional) dataName. This makes the class cleaner to mock.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/GCSBatchTableWriter.java
```diff
@@ -23,15 +23,15 @@
...
 import org.apache.kafka.connect.errors.ConnectException;
 
+import org.apache.kafka.connect.sink.SinkRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
criccomini, line 34: Wildcard again

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
```diff
@@ -100,11 +98,11 @@ public GCSToBQWriter(Storage storage,
    * @param blobName the name of the GCS blob to write.
    * @throws InterruptedException if interrupted.
    */
-  public void writeRows(List<RowToInsert> rows,
+  // writeRows -> needs to be changed
```
criccomini, line 101: ?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
```diff
@@ -196,10 +194,10 @@ private void waitRandomTime() throws InterruptedException {
     Thread.sleep(retryWaitMs + random.nextInt(WAIT_MAX_JITTER));
   }
 
-  private void attemptTableCreate(TableId tableId, String topic) {
+  private void attemptTableCreate(TableId tableId, Set<SinkRecord> records) {
     try {
-      schemaManager.createTable(tableId, topic);
-      logger.info("Table {} does not exist, auto-created table for topic {}", tableId, topic);
+      schemaManager.createTable(tableId, records);
+      logger.info("Table {} does not exist, auto-created table ", tableId);
```
criccomini, line 200: Log message should come before `schemaManager.createTable`

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
```diff
@@ -30,9 +30,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
criccomini, line 33: Wildcard

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/BigQueryWriter.java
```diff
@@ -25,14 +25,11 @@
...
+import org.apache.kafka.connect.sink.SinkRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
-import java.util.Set;
+import java.util.*;
```
criccomini, line 32: Wildcard

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -30,12 +30,11 @@
...
 import com.wepay.kafka.connect.bigquery.utils.PartitionedTableId;
 
+import org.apache.kafka.connect.sink.SinkRecord;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
criccomini, line 37: Wildcard

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/batch/TableWriter.java
```diff
@@ -72,16 +66,21 @@ public void run() {
     int successCount = 0;
     int failureCount = 0;
 
+    List<Map.Entry<SinkRecord, RowToInsert>> rowsList = new ArrayList<>(rows.entrySet());
     try {
       while (currentIndex < rows.size()) {
-        List<RowToInsert> currentBatch =
-            rows.subList(currentIndex, Math.min(currentIndex + currentBatchSize, rows.size()));
+        List<Map.Entry<SinkRecord,RowToInsert>> currentBatchList =
```
criccomini, line 72: Nit: `SinkRecord, RowToInsert` (space)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
```diff
@@ -182,7 +180,7 @@ private boolean onlyContainsInvalidSchemaErrors(Map<Long, List<BigQueryError>> e
     boolean invalidSchemaError = false;
     for (List<BigQueryError> errorList : errors.values()) {
       for (BigQueryError error : errorList) {
-        if (error.getReason().equals("invalid") && error.getMessage().contains("no such field")) {
+        if (error.getReason().equals("invalid") && (error.getMessage().contains("no such field") || error.getMessage().contains("Missing required field"))) {
```
criccomini, line 183: I don't think this is going to work. These error.getReason/error.getMessage calls are to handle errors coming from BigQuery itself, not from our code.

Are you trying to deal with exceptions from https://github.com/wepay/kafka-connect-bigquery/pull/291/files#diff-925c66dfbab22d569e960434966e6f75R145 ?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
C0urante: Ah, gotcha. So really we're replacing `autoUpdateSchemas` with a combination of `addNewBigQueryFields` and `changeRequiredFieldsToNullable`, right? That should cover the only two schema evolutions based on the [BigQuery docs](https://cloud.google.com/bigquery/docs/managing-table-schemas).

Will the unionization logic be configurable? Think we might want to make it toggle-able and even off by default, since it's a bit of a footgun if you accidentally send even just one bad record to the connector (a bunch of unnecessary columns will get added to your table and BigQuery doesn't allow deletion of columns, so you're pretty much stuck with them unless you want to delete and then recreate the table).

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
stoynov96: We could validate this in the `validate` method probably

kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/BigQueryConnectorIntegrationTest.java
```diff
@@ -48,10 +48,7 @@
 import java.time.LocalDate;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
-import java.util.ArrayList;
```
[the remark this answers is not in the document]
sahithi03: Will change this. Updated IDE settings.

kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/utils/BucketClearer.java
```diff
@@ -42,12 +43,13 @@ public static void main(String[] args) {
...
-    if (gcs.delete(bucketName)) {
-      logger.info("Bucket {} deleted successfully", bucketName);
-    } else {
-      logger.info("Bucket {} does not exist", bucketName);
+    Bucket bucket = gcs.get(bucketName);
+    if (bucket != null) {
+      logger.info("Deleting objects in the Bucket {}", bucketName);
+      for (Blob blob: bucket.list().iterateAll()) {
+        gcs.delete(blob.getBlobId());
+      }
```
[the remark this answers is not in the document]
sahithi03: Ah! Will add it.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
sahithi03: Will update accordingly

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
sahithi03: Yes! Will update the doc.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
sahithi03: No, was not trying to deal with exceptions from the code. `Invalid: Missing Required field `is the error message from google when we try to insert a value with NULLABLE mode into a field with REQUIRED mode. I will change the message in the exception in [https://github.com/wepay/kafka-connect-bigquery/pull/291/files#diff-925c66dfbab22d569e960434966e6f75R145](url), since it can be confusing.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
sahithi03: oops, left this here by mistake. Will remove it.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/GCSToBQWriter.java
sahithi03: will update it.

kcbq-connector/test/resources/connector-template.properties
sahithi03: No, this will not have any effect on the records. The regex was just meant to be an example/placeholder :)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/utils/SinkRecordConverter.java
sahithi03: Will update accordingly.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -67,32 +68,90 @@ public SchemaManager(
...
+      currentSchema = unionizeSchemas(currentSchema, schemas.get(i));
+    }
+    return currentSchema;
+  }
+  private com.google.cloud.bigquery.Schema unionizeSchemas(com.google.cloud.bigquery.Schema currentSchema, com.google.cloud.bigquery.Schema newSchema){
+    Map<String, Field> currentFields = new HashMap<>();
+    Map<String, Field> newFields = new HashMap<>();
+    FieldList currentSchemaFieldList = currentSchema.getFields();
+    FieldList newSchemaFieldList = newSchema.getFields();
+    for (Field field: currentSchemaFieldList) {
```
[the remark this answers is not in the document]
sahithi03: Nice! Will make it more compact

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -67,32 +68,90 @@ public SchemaManager(
...
+      currentFields.put(field.getName(), field);
+    }
+    for (Field field: newSchemaFieldList) {
+      newFields.put(field.getName(), field);
+    }
+    for (Map.Entry<String, Field> entry: newFields.entrySet()) {
+      if (!currentFields.containsKey(entry.getKey())) {
+        if (autoAddNewFields) {
+          currentFields.put(entry.getKey(), entry.getValue().toBuilder().setMode(Field.Mode.NULLABLE).build());
```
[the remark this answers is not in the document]
sahithi03: Yes! This is better than adding the new field as `NULLABLE` by default. Will change this.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -67,32 +68,90 @@ public SchemaManager(
...
+        } else {
+          throw new BigQueryConnectException("No field with the name " + entry.getKey() );
+        }
+      } else {
+        if (currentFields.get(entry.getKey()).getType() != newFields.get(entry.getKey()).getType()) {
```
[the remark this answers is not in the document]
sahithi03: Should be the latter. Will update this.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -67,32 +68,90 @@ public SchemaManager(
...
+          throw new BigQueryConnectException("Incompatible Schema: Field Type Mismatch");
+        }
+        if (currentFields.get(entry.getKey()).getMode() == Field.Mode.REQUIRED && newFields.get(entry.getKey()).getMode() == Field.Mode.NULLABLE){
```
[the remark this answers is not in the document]
sahithi03: Since modes are enums, should I switch to `equals` because `getMode() `would be null-safe?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -67,32 +68,90 @@ public SchemaManager(
...
    * Create a new table in BigQuery.
    * @param table The BigQuery table to create.
-   * @param topic The Kafka topic used to determine the schema.
+   * @param records The sink records used to determine the schema.
    */
-  public void createTable(TableId table, String topic) {
-    Schema kafkaValueSchema = schemaRetriever.retrieveSchema(table, topic, KafkaSchemaRecordType.VALUE);
-    Schema kafkaKeySchema = kafkaKeyFieldName.isPresent() ? schemaRetriever.retrieveSchema(table, topic, KafkaSchemaRecordType.KEY) : null;
-    bigQuery.create(constructTableInfo(table, kafkaKeySchema, kafkaValueSchema));
+  public void createTable(TableId table, Set<SinkRecord> records) {
```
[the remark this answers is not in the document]
sahithi03: I tried to avoid looping through the Records twice.(Once for getting the unionized schemas and again for the table description). But, this is much cleaner and readable. Will split it up the method.

kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/BigQueryConnectorIntegrationTest.java
```diff
@@ -176,8 +173,15 @@ private Object convertField(Field fieldSchema, FieldValue field) {
     List<Object> result = new ArrayList<>();
     assert (rowSchema.size() == row.size());
 
+    for (int i=0; i < rowSchema.size(); i++) {
+      if (rowSchema.get(i).getName().equals("row")) {
```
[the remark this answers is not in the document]
sahithi03: The` schema unionization `logic creates fields in the BigQuery table which might not follow the same order as the fields in the `schema.json` file. The integration tests compare the actual and expected rows based on `testRow.get(0`) which is supposed to be the `“row”` field. But since the fields would now be shuffled, I tried to add the `“row”` field values first in the `testRow` list to make sure that it is the first value. Then in the subsequent loop skipped the "row" field values, so that these are not entered twice in the list. But I would want to know if there's a better way.

kcbq-connector/quickstart/properties/connector.properties
```diff
@@ -22,18 +22,23 @@ sanitizeTopics=true
...
 
 bufferSize=100000
 maxWriteSize=10000
 tableWriteWait=1000
 
+transforms=RegexTransformation
+transforms.RegexTransformation.type=org.apache.kafka.connect.transforms.RegexRouter
+# .* is a placeholder for the actual regex
```
stoynov96, line 33: Maybe rephrase as something like "A placeholder regex router SMT that does nothing. Replace with relevant transformation SMT"

kcbq-connector/quickstart/properties/connector.properties
```diff
@@ -22,18 +22,23 @@ sanitizeTopics=true
...
+
 ########################################### Fill me in! ###########################################
 # The name of the BigQuery project to write to
 project=
 # The name of the BigQuery dataset to write to (leave the '.*=' at the beginning, enter your
 # dataset after it)
 datasets=.*=
 # The location of a BigQuery service account JSON key file
-keyfile=
+keyfile=
```
stoynov96, line 44: nit: newline at the end of the file

kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/BigQueryConnectorIntegrationTest.java
```diff
@@ -50,9 +50,11 @@
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Comparator;
 import java.util.List;
 import java.util.Properties;
 
+
```
stoynov96, line 57: nit: redundant new line

kcbq-connector/src/integration-test/java/com/wepay/kafka/connect/bigquery/it/BigQueryConnectorIntegrationTest.java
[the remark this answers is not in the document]
stoynov96: Not sure how generalized indexOf() can be in java, can we use this to make this cleaner?

Edit: looks like we can't specify a comparator for it actually

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -133,15 +133,35 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
     topicPartitionManager.resumeAll();
   }
 
+  private void maybeEnsureExistingTable(TableId table) {
+    BigQuery bigQuery = getBigQuery();
+    if (bigQuery.getTable(table) == null && !config.getBoolean(config.TABLE_CREATE_CONFIG)) {
+      logger.warn("You may want to enable auto table creation by setting {}=true in the properties file",
```
stoynov96, line 139: Can we append this warning to the exception message instead?

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/BigQuerySinkTask.java
```diff
@@ -133,15 +133,35 @@ public void flush(Map<TopicPartition, OffsetAndMetadata> offsets) {
...
+    String[] smtReplacement = record.topic().split(":");
+
+    if (smtReplacement.length == 2) {
+      dataset = smtReplacement[0];
+      tableName = smtReplacement[1];
+    } else if (smtReplacement.length == 1) {
+      tableName = smtReplacement[0];
+    } else {
+      throw new ConfigException("Incorrect regex replacement format. " +
+              "SMT replacement should either follow <dataset>:<tableName> format or replace by <tableName> only.");
```
stoynov96, line 157: This message could be a bit unclear. I think it'd be clearer like
`SMT replacement should either produce the <dataset>:<tableName> format or just the <tablename> format`.

kcbq-connector/test/resources/connector-template.properties
C0urante: Ah, gotcha. In that case, can we do one of the following (no string preference on my part for/against any of them):
- Add a comment explaining what this is and how someone might alter it to achieve table routing behavior
- Remove it
- Alter it to provide a more practical example (such as appending or stripping a prefix or suffix, or redirecting all records from one topic to a different topic but leaving all others unaffected)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
criccomini: Yea, agree. It should be configurable, and off by default.

kcbq-connector/quickstart/properties/connector.properties
criccomini: Yes, I think so.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
```diff
@@ -67,32 +83,106 @@ public SchemaManager(
...
     logger.info("Attempting to update table `{}` with schema {}",
-        table, tableInfo.getDefinition().getSchema());
+            table, tableInfo.getDefinition().getSchema());
     bigQuery.update(tableInfo);
   }
 
-  // package private for testing.
-  TableInfo constructTableInfo(TableId table, Schema kafkaKeySchema, Schema kafkaValueSchema) {
-    com.google.cloud.bigquery.Schema bigQuerySchema = getBigQuerySchema(kafkaKeySchema, kafkaValueSchema);
+  private TableInfo getTableInfo(TableId table, Set<SinkRecord> records) {
```
criccomini, line 105: Could you add some Javadocs on these new private methods? I think they're sufficiently complicated to warrant some docs.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
criccomini: Hmm. Thought on this a bit more. I believe the update logic is configurable and off by default now.

We have `allowNewBigQueryFields` and `allowBigQueryRequiredFieldRelaxation` now. These both default to false. I think these effectively covers the unionization logic, since the if is:

```
      if (!currentFields.containsKey(entry.getKey())) {
        if (allowNewBQFields && (entry.getValue().getMode().equals(Field.Mode.NULLABLE)
                || (entry.getValue().getMode().equals(Field.Mode.REQUIRED) && allowBQRequiredFieldRelaxation))) {
          ...
      } else {
        if (currentFields.get(entry.getKey()).getMode().equals(Field.Mode.REQUIRED) && newFields.get(entry.getKey()).getMode().equals(Field.Mode.NULLABLE)) {
          if (allowBQRequiredFieldRelaxation) {
            currentFields.put(entry.getKey(), entry.getValue().toBuilder().setMode(Field.Mode.NULLABLE).build());
            ...
```

Since both default to false, schemas won't get updated by default.

kcbq-connector/test/resources/connector-template.properties
criccomini: I vote for:

> Alter it to provide a more practical example (such as appending or stripping a prefix or suffix, or redirecting all records from one topic to a different topic but leaving all others unaffected)

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/write/row/AdaptiveBigQueryWriter.java
criccomini: Ahh! Gotcha.

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/SchemaManager.java
sahithi03: Sure. Will update.

kcbq-connector/quickstart/properties/connector.properties
```diff
@@ -39,6 +41,6 @@ transforms.RegexTransformation.replacement=$0
 project=
 # The name of the BigQuery dataset to write to (leave the '.*=' at the beginning, enter your
```
stoynov96, line 42: This comment seems to be outdated now. We should update it

sahithi03: Will update

kcbq-connector/src/main/java/com/wepay/kafka/connect/bigquery/config/BigQuerySinkConfig.java
C0urante: @criccomini @stoynov96 @sahithi03 sorry to dredge this up again, but I think we might want to reconsider some of the logic here.

The prior behavior of the connector was to basically send a single record's schema to BigQuery and let validation happen there; the [only permitted operations](https://cloud.google.com/bigquery/docs/managing-table-schemas) were (and still are) adding new columns to a table, and relaxing existing columns from `REQUIRED` to `NULLABLE`. This meant that it was possible to relax required fields to nullable, but only if there was a corresponding upstream schema change.

The new behavior of the connector still catches this case, but also automatically relaxes `REQUIRED` fields in the existing table schema to `NULLABLE` if they're missing from the most recent upstream schema.

This is risky, since it means that a single misplaced record with a completely disjoint schema from the existing table schema can cause permanent modifications to be made to the BigQuery table schema. Granted, this would require `allowNewBQFields` and `allowBQRequiredFieldRelaxation` to both be set to `true`, but it's not unreasonable for people to want to enable both with the expectation that they would cause the connector to act in the same way as it would have with `autoUpdateSchemas`.

I think we might still want to add a third config property, `allowSchemaUnionization`, that toggles the schema unionization behavior. If it's set to `false` and both `allowNewBQFields` and `allowBQRequiredFieldRelaxation` are set to `true`, then the prior behavior of the `autoUpdateSchemas` property should be preserved effectively for users who still want that.

WDYT?

